PR #12, opened 2021-11-23
Choose Directory for Chart Export

daminals:
User should be able to choose where function exports chart for organization and customizability, rather than use home directory of the computer


g-battaglia:
You can do it changing the value of this variable:
self.output_directory = self.homedir

daminals:
Yes in my pull request I added an option to change the variable self.output_directory by adding an optional parameter as the export directory

g-battaglia:
Ok, I merged the request with some minor changes, thank you very much!

PR #13, opened 2021-11-26
json_dump method: Minor tweaks re file output and location

daminals:
Changed the json_dump method so that the JSON output is in readable and parseable format, and added an ability to choose which directory to output the JSON file to

Screen Shot 2021-11-25 at 10 43 58 PM

Screen Shot 2021-11-25 at 10 44 17 PM


g-battaglia:
Hi, thank you for your contribution, I've merged the request!

Issue #15, opened 2021-12-06
Feature request: add a cache for geonames


mateuszmp:
I suggest calling an internal proxy when using geonames such as to only connect to it once for a given location even if there's many KrInstance objects for the same location. 
I didn't realize geonames was queried every time and created KrInstace objects in a loop with millions or iterations and received a time ban from geonames for effectively launching a denial-of-service against them.

mateuszmp:
Solved in PR: https://github.com/g-battaglia/kerykeion/pull/16

PR #16, opened 2021-12-07
Created a very simple proxy for the Geonames API

mateuszmp:
I created a simple proxy for the Geonames API as I have previously suggested in this issue: https://github.com/g-battaglia/kerykeion/issues/15

This change is backwards and forwards compatible.

The proxy works as follows:

- Previously urlopen() was called, now geonames_proxy() is called instead.
- This new function accepts two arguments: the operation (ie. search of place or timezone check) and the parameters to call the Geonames API with.
- geonames_proxy creates a directory (by default that's ```.kerykeion```) if it does not already exist. 
- It checks if there's a cache file available already
  - if yes then it will check if there's cached results of querying these parameters against the Geonames API
    - if there is then it just returns the past response for those same parameters
    - if there isn't then it queries the Geonames API, saves the response into the cache file, then returns it
- geonames_proxy will store JSON files containing parameters as keys and responses as values.

Features
- the ```.kerykeion``` directory can always be safely removed, as can any files within. This will simply prompt geonames_proxy to create it again and fill it with results
- all errors, information and warnings that I could foresee are logged using the ```logging``` module.
- this change allows for KrInstance objects to be initialized in a loop and at most 2 API calls will be made to Geonames per location (one for search and another for timezone)

Additional changes I've made:

- changed license url to use https
- changed some inconsistent indentation
- fixed an encoding issue which was making it impossible to pip install the module on Windows (that's a change in setup.py)
- wrapped the parsing of API response handling in a try-except to inform the user of a potential hourly rate limit hit against the Geonames API
- bumped the version number

### DIFF
--- a/kerykeion/geoname.py
+++ b/kerykeion/geoname.py
@@ -25,22 +25,124 @@ preset vars:
     maxRows=1
 """
 
-
-
-
 from urllib.request import urlopen
 from urllib.parse import urlencode
 from xml.dom.minidom import parseString
 from socket import timeout
 from urllib.error import HTTPError, URLError
+from os import mkdir, path, getcwd, remove
+import json
+import logging
+
+
+logging.basicConfig(
+    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s', level=logging.INFO)
+
+
 def _getText(nodelist):
-    """Internal function to return text from nodes
+    """"
+    Internal function to return text from nodes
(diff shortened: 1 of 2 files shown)


g-battaglia:
Hi, thank you very much for collaborating! I'll take a look on your pull request in the next days and if it's all ok I'll merge it!

PR #21, opened 2022-03-26
Support new language and small fixes.

alexsmithbr:
* Added: support to PT language.
* Fixed: output svg file name instead of dirname only.
* Fixed: readme.md renamed to README.md to make setup.py work properly.


g-battaglia:
Thank you very much for your contribution! It's grate!

PR #37, opened 2022-04-10
added offline mode

gaycodegal:
I wanted to do a computation without using online services. This project was the easiest way but did not support that, so I added it in. Should follow old code paths if online is true, which is the new default


g-battaglia (approved):
Hi, thank you! There are some things to fix (assert do not take parentheses) but I'll take care of that! Thank you very much for your contribution! 😊

PR #52, opened 2023-01-07
fix error and add french language

gallica-lupus:
- fixed an error on planet icons in the transits column : `xlink:href="# {self.planets_settings[i]["name"]}"` to `xlink:href="#{self.planets_settings[i]["name"]}"`
- add the translate for french (FR)


g-battaglia:
Thank you for your contribution!

PR #103, opened 2024-01-24
astrological_subject.py

must-be-rohit:
For Sidereal calculation update the mode from Fagan to Lahiri
  
      if self.zodiac_type == "Sidereal":
            self._iflag += swe.FLG_SIDEREAL
            #mode = "SIDM_FAGAN_BRADLEY"
            mode = "SIDM_LAHIRI"
            swe.set_sid_mode(getattr(swe, mode))

### DIFF
--- a/kerykeion/astrological_subject.py
+++ b/kerykeion/astrological_subject.py
@@ -341,7 +341,8 @@ class AstrologicalSubject:
 
         if self.zodiac_type == "Sidereal":
             self._iflag += swe.FLG_SIDEREAL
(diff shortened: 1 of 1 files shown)


g-battaglia:
Hi, it's ok but the best option should be to be able to chose what kind of sideral mode to use, not just default on SIDM_LAHIRI

g-battaglia:
```python
johnny = AstrologicalSubject("Johnny Depp", 1963, 6, 9, 0, 0, "Owensboro", "US", zodiac_type="Sidereal", sidereal_mode="LAHIRI")
```

PR #126, opened 2024-07-28
fix #125, disable fetching geonames if not online

### LINKED ISSUE #125: Topocentric offline usage error
when initializing `AstrologicalSubject`, setting `online=False` and `perspective_type="Topocentric"` at the same time will cause `KerykeionException`

hoishing:
only fetch geonames when online, lat, lng and timezone are set


g-battaglia:
Hi, thanks a lot, merged!

Issue #11, opened 2021-11-16
South Node calculation


meheszp:
Would it be feasible for the program to compute both South Node and return it in both individual and synastry aspects analysis?
Thank you very mucth!

g-battaglia:
So far there's no support for this, sorry!

fkostadinov:
Just created a pull request which should (hopefully) fix this issue. See here: https://github.com/g-battaglia/kerykeion/pull/134

PR #134, opened 2024-10-11
Added calculation of true and mean south nodes to Kerykeion

fkostadinov:
I really like this project, but I was very much missing an easy way to calculate the true and mean south nodes, which are very important in India, for example. So I decided to implement them myself based on your code.

As far as I could see the Swiss Ephemeris Library does not offer calculation of the south nodes directly. However, their degrees are really easy to calculate, as they are simply opposite the north nodes (180° apart). I added this calculation in _initialize_planets(). Also, I had to apply one or two more tricks there for the same reason and due to how the code was structured.

I was able to make most things work, but I was not sure how the test cases in tests/charts were supposed to be working with all the existing SVG files. I could not find a proper documentation on these SVG files. Therefore, I made all the test cases work - except the test/charts/test_charts.py one.

However, I tried to generate charts, and at least they seem fine from what I could see - except that now the font, because it's longer than before, might overlap with the chart. Haven't had time to fix that. What's missing is also a quick description  in README.md how to switch between true and mean nodes when generating the chart.

One caveat is that I had to rename "Mean Node" to "Mean North Node" and "True Node" to "True North Node" everywhere, to clearly distinguish them from the corresponding south nodes. The names are a bit long now, but for displaying they could potentially be shortened to something like "True N. Node" or "Tr. North Node". I haven't done this, though, did not want to interfere too much here.

Would love if you could merge that into your official code base, such that others can profit from this too!


fkostadinov:
FYI, this pull request would address this issue: https://github.com/g-battaglia/kerykeion/issues/11

fkostadinov:
One more comment: In the aspect table as well as in the chart there now exists always an opposition between the north and the south node. I think this makes sense, but obviously one further improvement could be to let a user consciously enable or disable the opposition. For example, similar like chiron and lilith it could be possible to consciously enable or disable individual celestial points in a list of points when constructing a chart. I haven't implemented this myself yet, but if there's interest, I could give it a try.

g-battaglia:
Hi, thank you so much! This is something I've been planning to do for a while, so I really appreciate your pull request. I'll review it thoroughly and aim to merge it as soon as possible. I've already taken a quick look and will likely rename the 'north node' to the old naming for backward compatibility. I'll update it to the full naming in version 5 of the library.

fkostadinov:
Ah, yes, the renaming for backward compatibility makes total sense.

PR #135, opened 2024-11-16
Fix SVG color format and saving as png

nasik353:
hello, 

thanks for this great library. here are the additions and modifications i made:

- fixed the svg color format so that the svg file can be converted to png and loaded with inkscape. 
- added a function makePNG to save the chart as png.
- reformat classic.css theme file to match the other themes.

cheers.

### DIFF
--- a/kerykeion/charts/kerykeion_chart_svg.py
+++ b/kerykeion/charts/kerykeion_chart_svg.py
@@ -37,13 +37,12 @@ from kerykeion.charts.charts_utils import (
     draw_planet_grid,
 )
 from kerykeion.charts.draw_planets import draw_planets # type: ignore
-from kerykeion.utilities import get_houses_list
+from kerykeion.utilities import get_houses_list, get_colors_css
 from pathlib import Path
 from scour.scour import scourString
 from string import Template
 from typing import Union, List, Literal
 from datetime import datetime
-
 
 class KerykeionChartSVG:
     """
@@ -611,6 +610,10 @@ class KerykeionChartSVG:
 
         else:
             template = template.replace('"', "'")
+
+        color_map = get_colors_css(self.color_style_tag)
+        for var, color in color_map.items():
+            template = template.replace(str(f"var({var})"), color)
 
         return template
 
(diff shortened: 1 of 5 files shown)


g-battaglia:
Hi, thank you for the pull request. I'll try to review it as soon as possible!

g-battaglia:
Okay, I was able to take a closer look at your pull request, and the code looks great. Thank you for also cleaning up the colors in the classic theme! The only thing I’d like to adjust a bit is the replacement of CSS variables with actual hex values. While it's an excellent idea to solve compatibility issues with most SVG editing applications and for PNG conversion, it does create a problem: these SVGs, once loaded directly on a page, can no longer be easily styled globally (here’s an example: https://www.kerykeion.net/birth-chart).

In the next few days, I’d like to implement a logic that makes this new feature backward-compatible. By default, it will be disabled, but a parameter (e.g., safe_css) could be provided to enable it. If the makePNG function is called directly, there would be a check for this class property, and if necessary, the template would be recompiled.

For now, I’ll merge everything into a separate branch. Thanks a lot for your contribution!

nasik353:
makes sense, i actually discovered a bug with my solution with transit charts, the png crops the couple aspects. 
anyway, i will follow this repo and pull once this is implemented.

g-battaglia:
@nasik353 I managed to test your integration more thoroughly. In summary, here are the elements that need some refinement:

- Some types of charts are not rendered correctly when converted to PNG (particularly transit charts and synastry charts), especially when certain placements are involved. This was the main issue that emerged in my previous tests and is a legacy of OpenAstro. It would be worth investigating how to handle these errors properly, as it's unclear why some placements work while others do not.
- The colors of the classic theme differ from the original ones, creating a problem of backward compatibility and causing confusion in terms of UX/UI for those who have already implemented the application. It should be adjusted to ensure full backward compatibility.

The dev/svg-to-png branch has been updated with the modification to enable/disable variable replacement. Let me know if you’d like to try completing this part of the integration and make another pull request so that we don’t overlap! In my opinion, the most complicated part is handling the placements; all chart types should be tested to ensure nothing breaks.

In any case, thank you so much for your contribution!

nasik353:
Thanks a lot for your time. I see now that there are many problems with this pr, I will dig deeper into the library to understand more how it works. I will also check this OpenAstro problem as it seems interesting to me.

PR #140, opened 2025-01-05
Patch to make geonames caching longer and changeable by user

tomshaffner:
It seems worthwhile to me to make the geonames cache timeout changeable by the user, and default to longer. It currently defaults to 24 hours, but given how rarely the underlying data changes, making it last for weeks or even months might be useful in many applications. It would be a backdoor way to download the data from the origin and thus avoid credit limits on the geonames API, but to do so only on-demand, and to also enable periodic checks for changes/updates in the underlying data.

This PR is one proposed way to implement it in the library directly, and to shift the default to 30 days. Feel free to tweak or make changes if this doesn't align with the paradigm you prefer; my point was to try and ease the work if this approach was good, but I don't care how it gets implemented if there's another better way.

Also of note, when I made this change locally I got 33 test failures afterwards, but for tests related to charts or test_astrological_subject. As best I can tell none of those failures relate to these changes though so I'm assuming they were failing prior to this? But may be worth a doublecheck.

For others with this issue prior to PR acceptance: In my _use_ of the library for the moment I'm implementing this myself by subclassing this way:

`from kerykeion.fetch_geonames import FetchGeonames
from requests_cache import CachedSession

class CustomFetchGeonames(FetchGeonames):
    def __init__(self, city_name: str, country_code: str, username: str = "century.boy", expire_after: int = 1814400):
        super().__init__(city_name, country_code, username)
        self.session = CachedSession(
            cache_name="cache/kerykeion_geonames_extended_cache",
            backend="sqlite",
            expire_after=expire_after,
        )`

### DIFF
--- a/kerykeion/astrological_subject.py
+++ b/kerykeion/astrological_subject.py
@@ -40,6 +40,7 @@ DEFAULT_SIDEREAL_MODE: SiderealMode = "FAGAN_BRADLEY"
 DEFAULT_HOUSES_SYSTEM_IDENTIFIER: HousesSystemIdentifier = "P"
 DEFAULT_ZODIAC_TYPE: ZodiacType = "Tropic"
 DEFAULT_PERSPECTIVE_TYPE: PerspectiveType = "Apparent Geocentric"
+DEFAULT_CACHE_EXPIRE_AFTER_DAYS = 30
 GEONAMES_DEFAULT_USERNAME_WARNING = (
     "\n********\n"
     "NO GEONAMES USERNAME SET!\n"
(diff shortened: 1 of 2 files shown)


g-battaglia:
Hi, thank you very much for the contribution, it seems like a great idea! Could you squash all the commits into a single one? Additionally, I would ask you to amend all the changes not directly related to the feature (e.g., I see that in several places, a whitespace has been added to lines with comments).

tomshaffner:
@g-battaglia, check the updated PR; think that should cover the requests.

PR #218, opened 2026-06-15
fix: birth-longitude LMT for pre-standardization dates

g-battaglia:
## Problem

Before standardized time zones existed (e.g. ~pre-1893 in Germany), each locality kept its own **Local Mean Time** based on longitude. For such dates the IANA tz database falls back to its initial **"LMT"** record, whose offset is the LMT of the zone's *reference meridian* (Berlin for `Europe/Berlin`), **not the birth city's**.

So historical charts were ~13 min off for births away from that meridian. Example — **Einstein** (Ulm, 9.99°E, `Europe/Berlin`):

| | before | after | astro.com |
|---|---|---|---|
| UTC | 10:37 (Berlin +00:53) | **10:50:02 (Ulm +00:39:58)** | 10:50 |
| ASC | Can 8.92° | **Can 11.64°** | Can 11°38' |
| MC | Pis 9.34° | **Pis 12.84°** | Pis 12°50' |
| Moon | Sag 14.40° | **Sag 14.53°** | Sag 14°31' |

Planets barely move in 13 min, but the angles/houses shifted ~3°.

## Fix

In `_calculate_time_conversions()`, when the localized datetime reports `tzname() == "LMT"` and a longitude is available, derive the UTC offset from the **birth longitude** (15° = 1 h, east ahead of UT), rounded to the whole second. This mirrors the existing, already-correct `_calculate_time_conversions_bce()` branch. Modern dates are untouched (their zone is not in the "LMT" period).

Also hardens `format_datetime_with_timezone()`: it built the `[±HH:MM]` label by slicing `strftime("%z")`, which mangles sub-minute offsets (`+003958` → `+0039:58`). It now derives the label from `utcoffset()` (backward-compatible for whole-minute offsets).

## Tests

Adds 3 **ephemeris-independent** regression tests (assert UTC/offset only, so they pass on DE440 and DE441):
- pre-standardization birth uses longitude LMT (Einstein → UTC 10:50:02, local +00:39:58);
- same zone, different longitudes (Berlin vs Ulm) → different offsets;
- modern date unaffected.

## ⚠️ Fixtures need regeneration on DE441

This change shifts the golden fixtures for all pre-standardization subjects (Einstein, Galileo 1564, Newton 1643, Enlightenment 1750, US 1776, Industrial 1850, …). They must be regenerated with `poe regenerate:all` on a **DE441-capable** machine — a de440-only box (range 1549–2650) drops the pre-1549 subjects and would corrupt the baseline, so the fixtures are intentionally **not** regenerated in this PR. The new regression tests guarantee correctness independently of the ephemeris kernel.

Verified the corrected output against astro.com to the arcminute (Einstein).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected historical pre-standardization timezone calculations to use the birth longitude’s LMT behavior, including consistent UTC/local timestamps and Julian day results.
  * Improved timezone label formatting for charts (cleaner `[±HH:MM]` offsets).

* **Tests**
  * Added regression coverage for pre-standardization and transition-boundary scenarios, including round-trip checks for ISO UTC inputs.
  * Updated the natal chart fixture output to reflect the corrected timezone handling.


g-battaglia:
@coderabbitai review

g-battaglia:
@coderabbitai review

g-battaglia:
Closing in favor of a fresh PR from the same branch so CodeRabbit performs a full review of the final state (CodeRabbit does incremental reviews and skips already-reviewed commits). No code changes — same branch `fix/historical-lmt-from-longitude`.

PR #219, opened 2026-06-15
fix: birth-longitude LMT for pre-standardization dates

g-battaglia:
> Supersedes #218 (closed) — reopened for a clean full CodeRabbit review of the final state. Same branch, no code changes since.

## Problem

Before standardized time zones existed (~pre-1893 in Germany, varies by country), each locality kept its own **Local Mean Time** based on longitude. For such dates the IANA tz database falls back to its initial **"LMT"** record, whose offset is the LMT of the zone's *reference meridian* (Berlin for `Europe/Berlin`), **not the birth city's**.

So historical charts were ~13 min off for births away from that meridian. Example — **Einstein** (Ulm, 9.99°E, `Europe/Berlin`):

| | before | after | astro.com |
|---|---|---|---|
| UTC | 10:37 (Berlin +00:53) | **10:50:02 (Ulm +00:39:58)** | 10:50 |
| ASC | Can 8.92° | **Can 11.64°** | Can 11°38' |
| MC | Pis 9.34° | **Pis 12.84°** | Pis 12°50' |
| Moon | Sag 14.40° | **Sag 14.53°** | Sag 14°31' |

Planets barely move in 13 min, but the angles/houses shifted ~3°.

## Fix

1. **`_calculate_time_conversions`** — when the localized datetime reports `tzname() == "LMT"` and a longitude is available, derive the UTC offset from the **birth longitude** (15° = 1 h, east ahead of UT), rounded to the whole second. Mirrors the existing, already-correct BCE branch. Modern dates are untouched (their zone is not in the "LMT" period).

2. **`from_iso_utc_time`** — for pre-standardization instants it converts UTC→local with the same birth-longitude LMT, and passes the resolved offset **explicitly** (internal `_lmt_offset_seconds`) to `from_birth_data`. This prevents a double-interpretation, and is robust when the longitude wall time lands just past an IANA LMT→standard **transition boundary** (e.g. `Africa/Nairobi` 1908-04-30T21:20Z, lng 41.9) — which would otherwise be re-localized as standard time and trip the round-trip guard.

3. **`format_datetime_with_timezone`** — derive the `[±HH:MM]` label from `utcoffset()` instead of slicing `strftime("%z")`, which mangled sub-minute offsets (`+003958` → `+0039:58`). Backward-compatible for whole-minute offsets.

## Tests & fixtures

- New **ephemeris-independent** regression tests (assert UTC/offset only): longitude LMT for `from_birth_data`, longitude sensitivity (Berlin vs Ulm), modern pass-through, `from_iso_utc_time` round-trip, and the Nairobi transition-boundary round-trip.
- Regenerated the **Einstein golden report** fixture and the **8 pre-standardization SVG baselines** for the intentional output change. These are kernel-agnostic at report/SVG precision (only historical subjects change; modern fixtures regenerate byte-identical).
- Curated suite `poe test:core` is **green** (3857 passed).

## ⚠️ High-precision data fixtures need regeneration on DE441

The full-precision historical position/aspect fixtures (`tests/data/expected_positions*.py`, `expected_aspects.py`, …) shift with this change and must be regenerated with `poe regenerate:all` on a **DE441-capable** machine. This dev box has `de440.bsp` (range 1549–2650), which drops pre-1549 subjects and would swap de441→de440 for every value, so those fixtures are intentionally **not** regenerated here. The ephemeris-independent regression tests guarantee correctness regardless of kernel; verified against astro.com to the arcminute (Einstein).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of historical astrological charts by correctly calculating local time from birth longitude for pre-standardization dates.
  * Enhanced timezone offset formatting in chart outputs.

* **Tests**
  * Added comprehensive regression tests for pre-standardization timezone handling and historical birth date calculations.


g-battaglia:
@coderabbitai review

g-battaglia:
Addressed in 90fde379. `lmt_offset_seconds` is now explicitly popped from `calc_data` right after `_calculate_time_conversions` consumes it, mirroring the existing `_active_fixed_stars` cleanup — so the model input no longer carries transient state and we don't rely on Pydantic's `extra='ignore'`.

Left `is_dst` untouched: it's a pre-existing extra key (set well before this PR), so popping it is out of scope here.

Verified: `tests/core/test_astrological_subject.py` 110 passed; `model_dump()` no longer contains `lmt_offset_seconds`; Einstein still resolves to UTC 1879-03-14T10:50:03 (local +00:39:57).

PR #223, opened 2026-06-21
fix: ultra-review v6 — 24 verified correctness, security & robustness fixes

g-battaglia:
## Ultra-review of the full v6 source

Two rounds. **Round 1:** 8 parallel agents reviewed every line; **every** finding was then re-verified by an adversarial subagent (90 findings → 56 confirmed, 31 false positives, 2 intentional). **Round 2:** 10 fresh finders on under-covered areas (translations 1779 lines, chart_drawer 5195-line internals, collision math, data-table consistency) + adversarial verification.

This PR applies **~38 verified fixes across 11 focused commits**. Full suite: **9227 passed · 0 failed · 1407 skipped**, `ruff` clean (also repairs a pre-existing float-boundary failure in `test_future_2050_snapshot`).

### Security / robustness
- `fetch_geonames`: HTTP timeouts (was unbounded → DoS); stop leaking the GeoNames `username` (credential) into DEBUG logs; broaden cache filter except.

### FastAPI / schema integrity (`get_type_hints()` verified intact)
- `Optional`-but-required fields get `default=None`; deprecated `kr_types` shim delegates `__all__` to `kerykeion.schemas.__all__`.

### Correctness
- `get_planet_house` span-agnostic (non-quadrant systems >180°); `format_ancient_iso` + `julian_to_datetime` round/carry (was ~1s off BCE, latent `seconds==60` ValueError); `degreeSum` non-negative; report element/quality % reuse model integers; fixed-star magnitude column; void-of-course dedup; reuse/drop redundant ephemeris calls; `planetary_nodes` raises on all-fail; stop fabricating White_Moon as Lilith+180°; `_get_zodiac_info` guards `sidereal_mode=None`.
- Plus: composite `__eq__`, validate `KERYKEION_LEB_MODE`, `HOUSE_SYSTEM_EQUAL "E"→"A"`, escape `kr:` SVG attrs, copy-on-store, serializer `.get()`+house numbers, document `backwards`, pytz `UnknownTimeZoneError`.

### i18n
- IT `"houses"` was "Domificazione" (== "domification") → "Case"; FR `transit_name` used "transition" → "Au moment du transit".

### Documentation (semantically-loaded; behavioural fix left to maintainer)
- `phase_angle` = Sun–Moon elongation; Gauquelin sector for ephemeris-less points is an ecliptic approximation; `axis_orb_limit` is a deliberate dual-chart no-op.

### Deferred for the maintainer (confirmed, but NOT applied — need a decision / canonical-kernel regen)
- **Axis-cusp colors off-by-one** (MEDIUM, confidence 0.97): `chart_drawer.py:3819-3822` & `3929-3932` use `planets_settings[12..15]` (Chiron, Ascendant, Medium_Coeli, Descendant) where they should use `[13..16]` (Ascendant, Medium_Coeli, Descendant, Imum_Coeli) — the ASC cusp gets Chiron's colour, MC gets ASC's, etc. The 1-line fix is trivial, but it changes rendered SVG and requires regenerating **all** chart baselines; on a short-range dev kernel some baselines (ancient Rome) can't regenerate, so this should be applied + regenerated on the full-DE441 kernel in one step.
- **Translation completeness**: 11 celestial points (Interpolated_Lilith, Mean/True_Priapus, 8 Uranian) and 11 chart-label keys (point, transit, transit_point, house_position/natal_house, progression*, synastry_chart) are missing from the non-EN language blocks → they leak the English/Latin default on non-EN charts. Needs accurate native translations (CN/RU/HI/…).
- Earlier deferred: local-space ecliptic latitude=0; `axis_orb_limit` actually filtering dual charts (doctrine); MoonPhaseSunInfo rise/set int/str convention (public-API types); almuten participating triplicity; plus low-value cosmetics.

### Follow-up review (round 2 — applied in this PR)
A second medium-effort review pass over this PR's own diff surfaced six items, all now fixed + regression-tested (suite: **9258 passed**, `ruff` clean):
- **`convert_decimal_to_degree_string`**: the `"3"` format could emit an invalid `60"` near a degree boundary (`10.99997°` → `10°59'60"`). Now rounds the total and carries via `divmod` → `11°00'00"`. Formats `"1"`/`"2"` keep floor semantics; every input still formats the same rounded total, so no valid baseline changes.
- **Gauquelin sector midpoint** (previously deferred): confirmed the old formula was **180° off** (sector numbers placed opposite the sector). The corrected midpoint was already in the diff and is now covered by a regression test. There are no Gauquelin SVG baselines (the label angle is computed, not snapshotted), so no baseline regen is required.
- **`get_planet_house` / `is_point_between`**: the house lookup had inlined a second copy of the arc-containment math. `is_point_between` gained an `allow_reflex` flag and `get_planet_house` now reuses it — one tested primitive for quadrant and non-quadrant (>180° span) systems.
- **`format_ancient_iso`**: a float-noise hour within ~0.5 s of midnight was clamped to `23:59:59` of the *wrong* day. Now carries the rollover into the next day via a proleptic-Julian `_next_proleptic_julian_day` helper.
- **`day_length` formatting**: `report.py` (`H:MM`) and `context_serializer.py` (`str(timedelta)` → `H:MM:SS`) disagreed. Extracted a shared `format_timedelta_hhmm`; both surfaces now render `H:MM`.
- **`active_aspects` serialization**: `a.get('orb')` could render the literal string `None`. Now omits the `(orb)` suffix when orb is absent and skips nameless entries.

🤖 Generated with an adversarial multi-agent review + verification workflow ([Claude Code](https://claude.com/claude-code))

## Release Notes

* **New Features**
  * Added public API for computing Dorothean triplicity lords via `get_triplicity_lords()`.
  * Added progression-related language labels to the settings UI.

* **Bug Fixes**
  * Improved sidereal mode fallback and axis-orb filtering for dual-chart aspects.
  * Corrected ASC/MC/DSC/IC cusp coloring and Equal house-system identifier.
  * Improved White Moon handling when the ephemeris backend lacks support.
  * Moon phase outputs and XML serialization now use timezone-aware datetimes and proper day-length formatting.

* **Improvements**
  * Updated report percentage display to whole numbers; strengthened GeoNames, SVG escaping, and local-space ecliptic-latitude handling.


g-battaglia:
@coderabbitai review

---
_Generated by [Claude Code](https://claude.ai/code)_

g-battaglia:
Pushed follow-up fixes in 66f006ce addressing three confirmed findings from a second-pass review:

- **`utilities.py`** — `format_ancient_iso` could emit an invalid `:60` UTC-offset field (minutes rounded independently of hours); now carries into the hour via `divmod`.
- **`transits_time_range_factory.py`** — `axis_orb_limit` docstring still said dual-chart calcs ignore it, but transits run through the dual-chart path that now applies the filter; doc updated to match.
- **`fetch_geonames.py`** — debug-log `len()` assumed `geonames` was a list; a scalar payload raised an uncaught `TypeError`; guarded with an `isinstance(list)` check.

Added regression tests for the offset carry and the malformed-payload guard. Suite green, ruff clean.

@coderabbitai full review

g-battaglia:
@coderabbitai full review

New commit `3933cd2a` addresses a follow-up correctness bug found while reviewing this PR's own diff:

- **`relocated_chart_factory.py` (BCE path)**: this PR made `format_ancient_iso` round to the nearest second (with day carry), but the relocated BCE path still truncated the same `revjul` decimal hour with `int()`. Both derive from the same float, so the integer `hour`/`minute`/`seconds` fields and the stored ISO string diverged by up to ~1s — or a whole day at the midnight boundary (e.g. `dec_hour=23.9998889` → old int-fields `23:59:59` vs ISO `00:00:00` of the next day), violating the helper's documented invariant that the two agree. The integer fields now apply the same round + day-carry (reusing `_next_proleptic_julian_day`), passing the original un-carried date to `format_ancient_iso` to avoid a double day-carry.
- Updated the unit test that encoded the old truncation semantics and added a midnight-boundary regression test.
- Corrected the now-stale comment in `convert_decimal_to_degree_string` (only formats "1"/"2" floor; format "3" rounds).

Please pay particular attention to the day-carry interaction between the integer fields and `format_ancient_iso` (no double-carry), and the proleptic-Julian leap-year handling for negative astronomical years.

g-battaglia:
Chiusa e riaperta come PR identica per richiedere una nuova review completa a CodeRabbit.

Issue #10, opened 2021-10-30
SVG Encoding error


kit678:
Browser isn't able to render properly. Getting the following error:

This page contains the following errors:
error on line 18 at column 60: Encoding error
Below is a rendering of the page up to the first error.

________________________________
It renders properly up till birth date and time, so essentially renders 4 lines (including the name)

PR #231, opened 2026-07-02
v6.0.0 release-readiness: verified fixes from the deep review

g-battaglia:
## Summary

Implements every confirmed finding from the multi-agent deep review of `alpha/v6` (41 candidates → ~25 empirically confirmed), grouped into five independently-committed batches. Full suite green: **9,310 passed / 0 failed**; wheel builds and was smoke-tested from a clean venv.

### A — Blocking bugs (silently wrong data)
- **`nod_aps_ut` argument order** (`fix(planetary-nodes)`): `method="mean"` silently returned *osculating* nodes on both backends (Moon mean asc node 123.95° instead of 125.04° @ J2000). The regression test made the same swapped call; it now asserts `mean != osculating`.
- **Gauquelin sectors** (`fix(charts)`): `houses_ex2(b'G')` cusps *descend*, but the drawers assumed ascending — classic labels sat exactly 180° off, the modern overlay was mirrored (`rotate(+angle)` vs `rotate(-angle)` everywhere else), fallbacks were anchored at 0° Aries instead of the ASC, and hit-area wedges had inverted sweep flags. New invariant tests lock rotation-sign parity with the house ring at the ASC.
- **BCE Davison composites** (`fix(composite)`): midpoints decomposed proleptic-Gregorian while the BCE birth path re-encodes Julian+LMT — charts landed ~74 h off at year −100. Now an exact inverse (round-trip < 1 s), unit-tested with pure `julday`/`revjul` math in every test tier.
- **Planetocentric center body** (`fix(subject)`): a Marscentric chart carried the *geocentric* Mars labelled as planetocentric. The center body is now excluded with warnings at three levels (active-points resolution, calc loop, Arabic-Part prerequisites).

### B — Release hygiene
- Removed the 8 pre-b1 deprecated aliases + `kerykeion/_deprecation.py`, honoring the "removed in 6.0.0 stable" promise in their warnings.
- `libephemeris` pin relaxed to `>=3.0.0rc1,<4` (**TODO before tagging: bump the floor once 3.0.0 stable is on PyPI**); README documents the bundled 1849–2150 data range and tier downloads.
- swisseph backend auto-detects `~/.kerykeion/sweph` (the setup script's download target, now a shared constant) — previously the downloaded files were ignored without `KERYKEION_EPHE_PATH`, silently dropping Chiron and fixed stars.
- `next_return_from_year` deprecation names its removal version (7.0.0).

### C — Minor correctness
Ante-CE weekday from the local date; heliacal BCE datestamps as `-0049`; sub-second precision in heliocentric-return/node-crossing charts; dedicated glyphs for the five Lilith/Priapus/apse points (were all the generic FixedStar star); fixed stars count in element/quality distributions again; `DualReturnChart` validates secondary points like other dual types. SVG baselines and report golden files regenerated.

### D — Performance (verified identical output)
Per-planet scan steps sized from station acceleration + bisection stopped at the output's 1-second granularity: 100-year full-planet scans go **354 s → 44 s** (ingresses) and **107 s → 10 s** (stations), with **4,963 ingresses and 1,824 stations event-identical** (timestamps within 10 ms). Mercury keeps the half-day step — its boundary excursions can last mere hours (locked by the 1970-01-04 double-crossing test).

### E — Consolidation before the API freeze
One shared `jd_to_iso_utc` (was 5 copies, 2 divergent day-boundary behaviors), one `utilities.wrap_180` (was 4 variants, 2 boundary conventions), planet-id maps derived from `POINT_NUMBER_MAP`, midpoint delegating to `circular_mean`.

## Before tagging 6.0.0
- [ ] Regenerate `tests/fixtures/natal_ancient_rome_all_report.txt` on a machine with the extended DE441 kernel (distribution totals changed; this machine only has the medium tier — the regen script skips it by design).
- [ ] Bump the `libephemeris` floor to `>=3.0.0,<4` when the stable release lands.
- [ ] The new BCE Davison/weekday integration tests only run under the extended tier.

## Test plan
- Full suite: 9,310 passed / 1,410 skipped (tier-gated) / 0 failed.
- `uv build` + install into a clean venv: chart renders, mean lunar node 125.041 @ J2000 from the installed wheel.
- Before/after event-parity comparison for the perf changes (1980–2080, all planets).

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

## Summary by CodeRabbit

* **New Features**
  * Added support for more fixed stars and additional chart symbols in reports and SVG charts.
  * Improved support for return, composite, progression, and heliacal outputs with clearer timestamps and chart labels.

* **Bug Fixes**
  * Fixed chart rendering, aspect detection, house assignment, transit sorting, and sunrise/sunset edge cases.
  * Improved handling of polar locations, BCE dates, timezone changes, and missing ephemeris data.

* **Documentation**
  * Updated setup and usage notes, including supported date ranges and ephemeris download guidance.

* **Breaking Changes**
  * Empty point filters are now treated as truly empty; use the new sentinel behavior for default points.


g-battaglia:
All 7 review findings addressed in 735d1cbf — full suite green (9,311 passed / 0 failed).

1. **Center-body emptying a non-empty `active_points`** → now raises `KerykeionException` instead of inverting into a full chart via the empty-list "no filter" semantics. Test updated (`test_only_center_requested_raises`).
2. **Star loop ignoring the caller's point filter + planet-grade fallback** → star inclusion is now **opt-in** (`include_fixed_stars=True`, passed by the chart data factory), so `calculate_element_points(..., ["sun","moon"], method="pure_count")` sums exactly 2.0 again; stars weigh `_FIXED_STAR_FALLBACK_WEIGHT = 0.2` when absent from the table, never the 1.0 point fallback. Both repros are regression tests now.
3. **`swisseph_setup` importing `ephemeris_backend`** → `DEFAULT_SWEPH_DOWNLOAD_DIR` moved to `settings/config_constants` (dependency-free); the setup script no longer touches the backend module. One correction to the repro, verified empirically: `python -m kerykeion.swisseph_setup` executes the parent package `__init__`, which imports the backend at line 65 — so with `KERYKEION_BACKEND=swisseph` exported and pyswisseph missing, the script fails identically **on alpha/v6 too** (`import kerykeion` alone reproduces it). The layering fix is still right; the bootstrap ordering (download first, export after) is what the script's own final instructions prescribe, and fixing the package-init eagerness is out of scope here.
4. **Third slug convention** → the star weight lookup uses the shared catalog slugger (new public `fixed_stars.catalog.star_slug`, strip + spaces/hyphens → underscores): `' Spica '` now resolves its 0.2 table weight; regression test included.
5. **`REQUIRED_SECONDARY_CHART_TYPES` ≡ `DOUBLE_CHART_TYPES`** → aliased, with a comment explaining why they must stay one name.
6. **`jd_to_iso_utc` sibling copies** → new shared `jd_to_ymd_hms(jd, cal)` core; `jd_to_iso_utc` is a thin formatter over it, the Davison decomposition reuses it (round+carry block deleted), and the heliacal datestamp is `jd_to_iso_utc(jd).split("T", 1)[0]` as suggested. `utilities._split_decimal_hour_with_carry`/`_assemble_ancient_iso` were left as-is deliberately: they operate on already-split components with a proleptic-Julian day carry (no JD in hand), a different input domain — merging them would give the component path a calendar parameter and a revjul dependency it deliberately avoids.
7. **`_MAX_SAMPLES` admission vs per-planet steps** → the guard now divides by the finest step actually requested, so slow-body-only scans get proportional headroom; the stale "~2700 years" comments are corrected (stations: ~38,000 years at 7d). The unreachable `.get(name, _SAMPLE_STEP_DAYS)` fallback stays as a defensive default for future bodies added to `_PLANET_IDS`.

Thanks also for the three verified-and-discarded candidates — the 800-year no-missed-ingress sweep near stations is a nice confirmation of the step calibration.

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

g-battaglia:
## Second review pass: 43-finding full-package review addressed

Independent re-verification (4 adversarial verifier agents with empirical repros) confirmed **42/43 findings** (F35 partially: the llms.txt block compiles but mixes factories — fixed anyway; one repro's mechanism corrected: the swisseph_setup bootstrap failure predates the PR at `kerykeion/__init__.py:65`). All fixes are in 8 batch commits (`dbf89f68..83645268`), suite green: **9,319 passed / 0 failed**, docs gate 75/75.

**Fixed (35 findings):**
- **B1 GeoNames** (F3/F4/F5/F16): OR-gating with fill-only-missing; explicit coords/tz never overwritten by the centroid; `from_current_time` resolves the target tz before capturing the instant; `KerykeionException` on failed lookups.
- **B2 draw_planets** (F1/F8/F22): angles classified by NAME (`CHART_ANGLE_NAMES` replaces the fossil 22<idx<27 window); the stray dual-chart tick block is gone; circular-aware indicator grouping. SVG baselines regenerated.
- **B3 aspects** (F2/F20/F21): adaptive lookahead (verified: Einstein's 0.12° Ascendant square flips from Separating to the correct Applying in the goldens); wrapped `diff`; closest-aspect classification.
- **B4 report/serializer** (F6/F15/F26): return labels from the actual `return_type`; XML positions driven by `active_points`; consistent "Transit" naming.
- **B5 crashes/wrong data** (F7/F9/F10/F11/F12/F13/F14/F17): UTC-stepped ephemeris series (monotonic across DST); default-seeded orb colors; Sun guard in the Discepolo score; polar clamp in relocation; midnight-sun re-pairing; heliacal backend-error handling **including the libephemeris jd=0.0 sentinel that emitted fake -4713 events**; nodes `method` validation; lunar-phase extreme branches.
- **B6 edge cases** (F18/F19/F23/F24/F25/F27/F28) and **B7 release items** (F32/F33/F34/F35 — note F34's failing snippet was introduced by this PR's own README note; now a text fence, gate green).
- **B8 safe cleanups** (F40/F41/F42/F43).

**Deliberately not done, with reasons in the commits:**
- F29: `libephemeris` floor bump — still blocked upstream (no 3.0.0 stable on PyPI yet; TODO in pyproject).
- F30/F31: Alpha classifier + consolidated 6.0.0 changelog section — tag-time steps.
- F36/F37/F39: consolidations with real behavioral caveats flagged by the review itself (BCE/microsecond format differences, azimuth-convention conversion).
- F38: the four dead `charts_utils` helpers are public documented API from the v6 PEP8 rename pass — kept.

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

g-battaglia:
## Third review pass addressed (commit 01cf06d3)

Suite green: **9,324 passed / 0 failed**. Findings 1–4 fixed, 5–6 deferred with reasons.

**1 — pure_count fixed stars weighed 0.2** ✅ Real bug from this PR's `include_fixed_stars=True`. `_prepare_weight_lookup` now returns a method-aware star fallback (0.2 in weighted, 1.0 in pure_count); threaded through the distribution helper. Verified: Sun+Moon+Regulus in pure_count now totals 3.0, weighted still adds 0.2. Regression test added.

**2 — DominantsFactory elemental inconsistent with ChartDataFactory** ✅ The elemental school now opts into fixed stars, restoring the documented "numbers match the chart distribution" invariant. Verified end-to-end that a star-bearing subject's dominants percentages equal `ChartDataModel.element_distribution`. Regression test added.

**3 — active_points=[] semantics change** ✅ Kept (None is the documented sentinel; `[]` as a real filter matches the timing factories), now documented as a **breaking change** in CHANGELOG, including the noted divergence from AstrologicalSubjectFactory.

**4 — center-body exclusion repeated in 3 sites** ✅ Took the deep fix you suggested: the guard now lives at the single chokepoint `_calculate_single_planet` (every planetary position is written there), so any current/future path inherits it. The three inline `_PLANETOCENTRIC_CENTERS.get` recomputations collapse to one `_planetocentric_center_names` helper; the redundant loop `continue` is removed. New test asserts a direct call for the center body stores nothing.

**5 — Gauquelin descending convention across ~7 sites** ⏸️ Deferred. The sites live in genuinely different coordinate spaces (classic offset space `-180 - i*10` vs modern wheel-angle space `360 - i*10`, plus per-space rotate-sign/sweep-flag/midpoint-arc), so a single source of truth is awkward and consolidating risks re-introducing the classic/modern mirror mismatch this PR just fixed. All sites are covered by the Gauquelin regression tests added earlier.

**6 — jd_to_ymd_hms divmod tail** ⏸️ Deferred, per your own caveat that it's not a drop-in (proleptic-Julian revjul vs the component splitter's calendar); only the divmod tail is shared, marginal.

Also fixed a pre-existing midnight-boundary flake in `test_from_current_time` (it compared a Europe/London chart against the host's local clock).

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

g-battaglia:
## Fourth-pass finding addressed (commit 29a8ec8d)

**Heliacal `next_heliacal_rising()` leaked a raw `ValueError`** on a no-event lookup — you're right, that's a genuine contract inconsistency I introduced with the B5 jd=0.0 guard. The multi-event `search_events` catches `_NO_EVENT_ERRORS` (ValueError + swisseph.Error), but the single-event method called `_find_event` unwrapped, so callers got a bare `ValueError` on libephemeris and a bare `swisseph.Error` on the swiss backend — bypassing the library's `KerykeionException` "no result" convention (nodes, retrograde, returns all raise it).

Fixed: `next_heliacal_rising` now normalizes both backend signals to `KerykeionException` with a documented `Raises` clause, so callers see one exception type regardless of backend. Regression test drives the libephemeris jd=0.0 sentinel via monkeypatch (astronomy-independent). Suite green.

Thanks for the four verified-and-refuted items too — agreed on all: the retrograde `prev_speed == 0.0` guard mirrors the existing `next_speed == 0.0` house style (sign-change is the real mechanism); the 0–180 `diff` and the midnight-rollover consolidation are intended/documented; the `nod_aps_ut` swap is the headline fix.

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

g-battaglia:
## Fifth-pass finding addressed (commit 789d313e)

**`from_current_time` fix was only partial** — you're right, and this is the pre-existing bug the PR meant to close. My B1 guard `if not tz_str and online and city and nation:` pre-resolved the timezone only when *both* city and nation were explicit. The default path (`city=None`) and city-without-nation both fell through to the else branch, captured `datetime.now()` as the host's naive wall clock, and `from_birth_data` then reinterpreted that host wall time in the resolved zone (Greenwich/GB → Europe/London, or the city under the default nation) — leaving the chart shifted by the full host-target offset.

Fixed: the timezone is now resolved for **every** online path with no tz_str, using the same `city or "Greenwich"` / `nation or "GB"` defaults `from_birth_data` applies, so the captured instant is always the true current moment. The else branch is now genuinely only reachable offline-without-tz_str, and its comment is corrected accordingly (you flagged it as misleading — it was). Regression test parametrized over the default and city-only paths.

Also corrected the `calculate_aspect_movement` lookahead comment you noted: it overstated that an exact aspect always reads Separating — two near-co-moving slow bodies can still read Static (behavior unchanged).

On the two persisted-output notes for the release notes — both are intended and already in CHANGELOG: `AspectModel.diff`/`kr:planetsdiff` now wrap-aware (359°/1° conjunction reads ~2, not 358), and the midnight-rollover datestamp in station/ingress (00:00:00 next day vs the old 23:59:59 clamp). Suite green (9,327 passed).

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

g-battaglia:
## Seventh-pass review addressed (commits 5d12246b, 1685cd6d)

Full-codebase audit (12 finders + 12 verifiers). I verified every finding independently before acting; only the clearly-contained, clearly-correct data/crash bugs were fixed in code, and the astrological/product judgment calls are documented (I asked for your call on those four — you were away, so I took the conservative reversible path and left the substantive method decisions open). Suite green (9,329 passed).

### Fixed (contained data/crash bugs)
- **sun_times** — mirror rescue for a real sunset just past local midnight (verified 69N/2026-05-19: was returning sunset/day_length/solar_noon = None with neither polar flag on a non-polar day). Genuine midnight-sun days still report polar_day.
- **get_kerykeion_point_from_degree** — a positive exactly 360.0 (swe_degnorm rounding on a non-pre-normalized point) used to raise and abort subject creation; all finite degrees now wrap, non-finite still raises.
- **EphemerisDataFactory** — `step <= 0` raises the documented ValueError, not a raw ZeroDivisionError.
- **transits** — events sorted by a signed extended-year key (BCE years were anti-chronological: `'-0400' < '-0500'` lexically).
- **context_serializer** — return subjects (PlanetReturnModel, a sibling not subclass) now emit their date/place block.
- **fetch_geonames** — `__get_timezone` calls `raise_for_status()`; both fetch methods surface a GeoNames `status` API error (invalid username/quota, returned as HTTP 200) clearly instead of a misleading "payload missing"; credential errors (code 10) become non-cacheable while the stable-negative "no result" (code 15) stays cached.

### Deliberately NOT changed
- **F7 (undersampling axes)** — reverted: axes are in DEFAULT_ACTIVE_POINTS and move ~360°/day, so no practical daily/hourly step resolves them; adding them fired the warning on every normal series, turning an actionable hint into noise.

### Judgment calls — documented, decision left to you
- **Converse directions** — the converse arc is the arithmetic complement (`360 - direct`), NOT the classical converse method. Flagged with a docstring warning + CHANGELOG "Known limitations" entry rather than removing the feature or attempting a delicate reimplementation in a release PR. **Your call: remove for 6.0.0, reimplement, or keep documented?**
- **Non-synastry dual-chart distributions** — element/quality computed from the natal subject only. Left as-is: a defensible "element balance is a property of the person, transit points are transient" reading. **Your call if you want Transit/Return/Progression to combine both subjects.**
- **BCE event calendar** — jd_to_iso_utc uses proleptic Gregorian (ISO 8601 mandates it); BCE natal dates use Julian (astro.com convention). Documented as a deliberate split — the helper is standards-correct, so left as-is.
- **Arabic-Part prerequisites** — Sun/Moon/Asc are calculated but stay out of `active_points` (internal dependencies). Left as-is: adding them would pull them into aspects/charts/reports the caller didn't request.

I can act on any of the four the moment you pick a direction.

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

g-battaglia:
## Eighth-pass review addressed (commit c2fbaf5e)

Full-codebase audit — 10 low/medium findings, no critical/high blockers on the ~82 reviewed files. Verified each independently before fixing. Suite green (9,332 passed).

### Fixed
- **#1 fixed-star-only active_points** — now raises (unified with the planetocentric center-body branch that already did), instead of silently inverting an emptied explicit list into a FULL chart. Also fixes the secondary symptom where `_calculate_arabic_part` extended the empty list mid-calc.
- **#2 context_serializer** — guarded the PlanetReturnModel date/place branch on its Optional lng/lat/iso fields (a hand-built model without them raised a TypeError from my own 7th-pass branch — thanks for catching it).
- **#3 secondary_progression** — the v6 enrichment flags (dignities/nakshatra/gauquelin/nutation/local-space/fixed-stars) are now propagated (inferred from the natal model), matching the docstring's "settings copied from the natal subject" promise and PlanetaryReturnFactory's parity.
- **#4 _get_lunar_phase_index** — `phase <= 0` raises (was returning Waxing Crescent).
- **#6 acg_factory** — reject an inverted lat_range (silently produced an empty grid).
- **#7 load_language_pair** — returns fresh deepcopies, never live references into LANGUAGE_SETTINGS (public entry point; mutation could poison process-wide defaults).
- **#8 is_out_of_bounds** — computed only for geocentric/topocentric perspectives; the declination-vs-obliquity test is physically meaningless for heliocentric/planetocentric frames (None there). Perspective fixtures regenerated.
- **#9 solar_arc** — the self-conjunction skip guard uses the same effective orb (base + per-pair extra_orb) as detection, so a per-pair widening can't reintroduce the tautological self-aspect.

### Left as-is with reasoning
- **#5** `to_context([])` producing an empty midpoints doc — a valid degenerate `list[MidpointModel]`, not a defect; forcing a TypeError would reject a legitimately-empty analysis.
- **#10** return-search topocentric altitude=0.0 — already a documented known limit.

### ⚠️ Coverage gap (worth acting on before tagging 6.0.0)
The reviewer's finder agents died on a spend limit, leaving two domains **entirely unreviewed**: `dominants/` + `dignities/` (almuten figuris, essential dignities, triplicity lords, scoring) and the astro-events group (eclipses, occultations, heliacal, lunations, moon_phase_details, planetary_phenomena, planetary_nodes, planetary_hours, void_of_course_moon) — ~22 files of sensitive astronomical math. The real residual risk for 6.0.0 is this missing coverage, not the findings above. Recommend a targeted review pass on just these two domains before the stable tag.

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

g-battaglia:
## Ninth pass — full coverage reached (104/104), commit e6ee5373

The two previously-uncovered domains are now fully reviewed and **clean** — no correctness bugs in the sensitive astronomical math (eclipses, lunar phases, void-of-course, planetary nodes/hours, dignities, almuten, prenatal syzygy). The one new item was a minor robustness-parity gap, now fixed:

- **PlanetaryPhenomenaFactory** caught per-planet backend errors and logged only, without the "all failed" guard `PlanetaryNodesFactory` has — an all-planets-failure returned an empty collection silently (indistinguishable from a valid "no phenomena"). It now raises `KerykeionException` for the all-failed case, matching the nodes factory. Regression test added.

Note: the review re-flagged the fixed-star-only `active_points` → full-chart issue as the top item to fix — that was **already fixed in the 8th pass** (commit c2fbaf5e), unified with the planetocentric center-body branch that raises.

### Status
All 104 files reviewed across nine passes; no critical/high blockers anywhere. Suite green. The remaining pre-tag TODOs (unchanged, tag-time/upstream): regenerate `natal_ancient_rome_all_report.txt` on the extended DE441 kernel, bump the `libephemeris` floor to `>=3.0.0` once the stable release lands, and the still-open astrological judgment calls from the 7th pass (converse-direction method, dual-chart distribution semantics) which are documented and awaiting your decision.

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

g-battaglia:
@coderabbitai review

g-battaglia:
## Addressed CodeRabbit's review (commit 4e5a58bb)

All three findings fixed — thanks @coderabbitai:

1. **`secondary_progression_factory.py` — enrichment inference gated on the Sun** (Functional, Minor). Correct catch: identical to the bug already fixed in `PlanetaryReturnFactory`. The Sun is `None` in a heliocentric natal (it's the excluded center body) or may be absent from a custom `active_points`, so keying dignities/nakshatra/local-space off `natal_subject.sun` silently dropped them. Now probes **any** populated standard body (same `_any_point_has` pattern as the return factory). Verified: a heliocentric natal with `calculate_dignities=True` now carries the enrichments into the progression; geocentric unchanged. Regression test added.

2. **`utilities.get_kerykeion_point_from_degree` — stale `Raises` docstring** (Maintainability, Minor). Fixed: the docstring now states it raises only for a non-finite degree; any finite degree (negative or ≥ 360) is wrapped into `[0, 360)`.

3. **CHANGELOG date-range claim vs README install note** (Maintainability, Minor). Fixed: the minor-body limitation note no longer implies 1550–2650 is available out of the box — clarified that the default `DE440s` install covers 1849–2150 and 1550–2650 needs the `medium` tier, matching the README's supported-date-range note.

Full suite green (9,376 passed). This branch has also been through 12 rounds of a multi-agent deep-validation review (whole-codebase, adversarially verified, with targeted 1:1-vs-Swiss precision checks); zero high/critical findings, all confirmed medium/low fixed.

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

g-battaglia:
## Addressed P2: normalize relocated longitude (commit 87f3d0f5)

Confirmed and fixed — thanks @coderabbitai. `relocate()` forwarded the raw `new_lng` to `houses_ex2`, so an un-normalized but equivalent longitude (`370` == 10° E, `-190` == 170° E) leaked a bare backend `CoordinateError`, inconsistent with the natal path which normalizes such inputs (round-8 fix). `relocate()` now wraps `new_lng` into `[-180, 180)` via `normalize_longitude` at the top, so it accepts wrapped longitudes and produces the same chart as the normalized value (verified: `lng=370` → stored `10.0`, ASC identical to `lng=10`; `-190` → `170`). `new_lat` already validated/clamped via `check_and_adjust_polar_latitude`. Regression tests added; full suite green (9,378 passed).

https://claude.ai/code/session_01DtZXJJQ2rCrHY6tgoESKVC

PR #234, opened 2026-07-20
fix: resolve timezones with zoneinfo, and stop clamping the polar ascendant

g-battaglia:
Three defects, all of them upstream or downstream of the astronomy rather than in it. Cross-checked against two independent references across five epochs, the ephemeris itself was already correct to under an arcsecond — every divergence we could measure traced back to one of these.

## The timezone floor and ceiling

`pytz` builds its transition table bounded by 32-bit `time_t`: first transition at `1901-12-13 20:45:52 UTC` (exactly −2³¹ s), last around 2037. Outside that window it does not extrapolate — it freezes the offset at the nearest known transition, so northern zones stay on standard time and southern zones stay on DST. **Eight of eight** DST-observing zones diverged, in both hemispheres.

Since the Ascendant advances 15°/hour, an hour of error is up to ~12° of Ascendant. A test subject literally named `Future Subject` (2100-07-04, New York) was being cast on EST instead of EDT — Ascendant off by **11.93°**, Midheaven by **13.84°**. That range is precisely the one the supported-year window was recently widened to cover.

`zoneinfo` fixes the low end for free. The predicate that refines a pre-standardization offset from the birth meridian keys on the record being named `LMT` — the database's way of saying it has nothing recorded — and that was already the right rule. It only misfired because `pytz`'s floor collapsed every pre-1901 instant onto the opening record. Named city mean times a zone genuinely kept (Bern, Moscow, Kyiv, Rome…) are recorded data and now survive, with their seconds intact where `pytz` rounded to whole minutes.

The `is_dst` → `fold` mapping was validated against `pytz` over 20 zones × every transition 1902–2037: 2 519 genuinely ambiguous instants, 5 038 decisions. The only divergences are in `Europe/Dublin`, a negative-DST zone where `pytz`'s own filter is inverted — there the new behaviour is the correct one.

## The polar fallback was reporting someone else's Ascendant

House systems undefined inside the polar circle were retried at a latitude clamped to ±66°, and **that retry's angles were reported as the subject's own**. So one place and instant produced different Ascendants depending on the house system — which cannot be true of a horizon intersection.

They now fall back to Porphyry at the real latitude: quadrant cusps, first cusp exactly on the Ascendant, angles exact. Midheaven was never affected and measurably does not move. The clamp survives only for Gauquelin, whose 36-cusp output shape admits no substitute — and it is now declared rather than silent, via a new `polar_house_fallbacks` list on the subject model.

The gate reads the backend's own polar threshold instead of a hardcoded 66°, because the real limit is 90° − obliquity and drops below 66° in high-obliquity epochs.

## Fixed stars had no provenance

The only points declaring neither `source` nor `precision_class`. Positions were already right — measured precession 49.85–50.57″/yr against ~50.3 expected — so this closes a contract gap, not an arithmetic one. Coverage and reviewed status stay `None` deliberately: the backend keys its coverage inventory by body id and has no star entries, so reporting a window would be an unbacked claim.

## Two latent defects, armed by the migration

Solar-noon arithmetic preserved wall clock instead of instant — harmless under a fixed-offset tzinfo, an hour's error across a transition once tzinfo became live. And an unguarded conversion at the year-1 boundary.

## Baselines regenerated, each with its cause

| Baseline | Δ | Cause |
|---|---|---|
| `industrial_1850` | ASC −0.0913°, MC −0.1191° | London adopted GMT in 1847, so 1850 is post-adoption; the old LMT was the 32-bit floor, not a record. Consistent with +31 s across Sun, Moon and MC. |
| `Future Subject` SVG | ASC −11.93°, MC −13.84° | EST → EDT. The headline case. |
| `Arctic` / `Antarctic` SVG | ASC +2.98° / +5.60° | MC, Sun and Moon move by **exactly 0.00000** — the signature of an observer no longer displaced to 66°. |
| `jung_1875` | score 32 → 28 | Kesswil kept Bern Mean Time 1853–1894; 450 s moved Jung's Moon 0.074°, pushing one opposition from 7.956° to 8.030° of orb, across the 8° limit. A knife-edge, pinned as such in the test docstring. |

**Deliberately unchanged, verified bit-identical:** `galileo_1564`, `newton_1643`, `enlightenment_1750`, `american_independence_1776`, `einstein_1879`.

## Dependency change

`tzdata` becomes a hard runtime dependency and `pytz` is gone. `zoneinfo` reads the host's timezone database and only falls back to the package — without the pin, two machines with differently-aged system data would compute different charts, and platforms without a system database would compute none.

## Verification

Full suite green: **9 982 passed, 1 387 skipped, 0 failed**, `ruff check` clean. Baseline measured on a pristine `HEAD` worktree rather than inferred. New regression tests cover post-2038 in both hemispheres, the six DST-transition cases with their exact messages, gap-before-ambiguity ordering, the 1893 adoption boundary, and the cross-house-system Ascendant invariant that fails on the old code above 66.5°.

---

https://claude.ai/code/session_01FrpVTaGnp5FcRax6a4appo

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timezone and daylight-saving handling, including historical dates, far-future dates, and transition-time calculations.
  * Corrected polar-latitude house calculations to preserve accurate angles and prevent Ascendant errors.
  * Fixed solar noon calculations across daylight-saving transitions.
  * Corrected select astrological position and relationship-score results.

* **New Features**
  * Added details describing when polar house-system fallbacks occur.
  * Fixed stars now include source and precision information consistently.


g-battaglia:
@coderabbitai review

Please pay particular attention to:

1. **The `is_dst` → `fold` mapping** in `utilities.py` (`localize_naive`). The rule is *`True` selects the larger `utcoffset()`, `False` the smaller* — deliberately not a hardcoded fold value, because the correspondence inverts between the fall-back (ambiguous) and spring-forward (gap) cases. The gap check must run **before** the ambiguity check, since a non-existent wall time also reports as ambiguous.

2. **The retained `tzname() == "LMT"` predicate** at the two sites in `astrological_subject_factory.py`. Keeping it means a birth elsewhere in a large zone before standardization inherits the reference city's recorded mean time rather than its own meridian. The reasoning is that `LMT` is the database's synthetic "nothing recorded" marker while `RMT`/`BMT`/`MMT`/`KMT` are civil times a zone genuinely kept — but it is a judgement call and worth a second opinion. The two predicates must stay identical to each other.

3. **The polar threshold gate** in `ephemeris_backend.py`. It now prefers the backend-reported threshold over a hardcoded ±66°, because the real limit is 90° − obliquity and drops below 66° in high-obliquity epochs. The no-op guard that prevents the swisseph backend's generic `Error` from being misread as a polar failure must not have been weakened.

4. **The `jung_1875` baseline move.** The score change depends on an orb crossing the 8° limit by 0.030°. Any sub-arcminute change flips it back — it is a regression pin, not a stable property, and the docstring says so.

g-battaglia:
Codex Review:

The patch improves the primary timezone and polar paths but still produces incorrect solar timing and DST disambiguation for valid historical inputs. It also leaves newly promised behavior broken on the supported swisseph backend.

Full review comments:

- [P2] Subtract UTC instants before deriving solar noon — /Users/giacomo/dev/kerykeion/kerykeion/moon_phase_details/factory.py:645-647
  When sunrise and sunset share the same `ZoneInfo` object but have different offsets, Python subtracts their wall-clock fields rather than their UTC instants. For Africa/Juba on 2000-01-15, where the offset changed from +02 to +03 at noon, this reports a 12:52 day instead of 11:52 and places solar noon 30 minutes late; convert both endpoints to UTC before calculating `half` and `day_length`.

- [P2] Preserve is_dst semantics for negative-DST folds — /Users/giacomo/dev/kerykeion/kerykeion/utilities.py:899-900
  The public parameter remains documented as a daylight-saving flag and previously followed `pytz` semantics, but selecting `True` solely as the larger offset reverses it when the DST reading has the smaller offset. For `Europe/Kyiv` at 1941-09-19 23:30, `is_dst=True` previously selected +02 but now selects +03, shifting the chart by one hour; resolve using DST status or expose a separate fold selector rather than redefining `is_dst`.

- [P2] Populate fixed-star provenance on the swisseph path — /Users/giacomo/dev/kerykeion/kerykeion/fixed_stars/discovery_factory.py:89-93
  With the supported `KERYKEION_BACKEND=swisseph` configuration, this guard never starts tracing, so `source` remains `None` and `_build_discovery_point` leaves both new provenance fields unset; the requested-star path has the same backend guard. This contradicts the newly advertised provenance contract and makes the added provenance tests fail whenever fixed stars are available through swisseph.

- [P2] Handle the moving polar threshold on swisseph — /Users/giacomo/dev/kerykeion/kerykeion/ephemeris_backend.py:339-354
  For ancient epochs where the polar threshold falls below 66°, swisseph raises its generic polar error without a `threshold` attribute at latitudes such as 65.9°. Since the fixed clamp leaves that latitude unchanged, this branch re-raises instead of substituting or clamping, so the newly added antiquity invariants fail under the supported swisseph backend and those charts remain uncast; derive the threshold from epoch obliquity or recognize the backend's polar diagnostic.

g-battaglia:
Both reviews verified against the code and against measurement before anything was touched. Two findings were real and are fixed; two were correct about the code but pointed at a limitation rather than a defect; one was declined.

## Fixed — elapsed time was measured on the wall clock

`moon_phase_details/factory.py`. Confirmed, and it is subtler than it looks. Python subtracts two aware datetimes by their **naive fields** whenever they share the same tzinfo *object*, never consulting the offsets — and `ZoneInfo` is cached, so two lookups of one zone hand back the same object. The shortcut is therefore the normal case, not an edge one.

```
Africa/Juba, 2000-01-15, offset changes +02 → +03 between the two
  sunset - sunrise                              = 12:00:00   ← what the code did
  sunset.astimezone(utc) - sunrise.astimezone(utc) = 11:00:00   ← actual
```

The earlier commit had converted the *addition* to instant space but left the subtraction that feeds it on the clock, so it was only half the correction. Both endpoints now cross to UTC before either operation.

## Fixed — `is_dst` was inferred from the offsets instead of asked of the zone

Confirmed on the exact case cited. Kyiv 1941-09-19 23:30 is genuinely ambiguous, and the two readings are Moscow time (+03:00, `dst=0`) and Central European Summer Time (+02:00). The daylight side is the **smaller** offset, so the offset-only rule handed `is_dst=True` the standard side.

The zone's own `dst()` now answers first; the clock decides only when both readings carry the same flag, which is what keeps negative-daylight-saving zones resolving the way a caller intends.

**This also exposed a flaw in my own earlier validation**, which had reported the offset rule as clean outside Ireland. It sampled around `pytz`'s transition instants, which are in **UTC**, while treating them as local wall times — so for zones with large offsets it was probing windows two or three hours away from the actual ambiguity. Re-measured correctly, sampling in local time every 5 minutes across 19 zones × every transition 1902-2037 (66 816 decisions):

| | divergences from the previous library | Dublin | Kyiv | Prague |
|---|---|---|---|---|
| offset only | 1657 | 1609 | **24** | 24 |
| zone first, then offset | 1633 | 1609 | **0** | 24 |

Dublin and Prague remain by design: there the other library books daylight saving with the opposite sign, which inverts the caller's meaning.

## Correct about the code, but a backend limitation rather than a defect

The two swisseph findings are accurate readings of the guards. They are not something the code can fix: that backend publishes no per-body provenance channel, and its polar diagnostic carries no measured threshold, so the epoch-dependent band below 66° cannot be recognised there at all. Fabricating a `source` or guessing a threshold would assert something unbacked.

What *was* wrong is that the new suites asserted those properties unconditionally, so they would have failed under `KERYKEION_BACKEND=swisseph`. They are now skipped on that backend with the reason stated. (Worth noting: I could not execute that backend here to confirm the failure directly — it needs `.se1` data files that are not present — so this rests on reading the guards, which are demonstrably absent.)

## Declined

The `ClassVar` annotation on `AMBIGUOUS_CASES`. `RUF012` is not in the project's selected rule set, and forcing it surfaces **six** occurrences of the same pattern in that one file. Annotating a single instance would leave the file less internally consistent than it is now; enabling the rule repo-wide is a separate change.

---

Full suite after the fixes: **9 986 passed, 1 387 skipped, 0 failed**, `ruff check` clean.

g-battaglia:
Codex Review (2):
• The ZoneInfo migration reverses fold resolution on TZif databases that encode negative daylight saving, producing one-hour errors on affected hosts. The polar substitution is numerically
  sound but emits a contradictory clamp log.

  Full review comments:

  - [P1] Handle negative-DST folds before trusting dst() — /Users/giacomo/dev/kerykeion/kerykeion/utilities.py:914-916
    When TZif encodes winter or Ramadan as negative DST, as PyPI tzdata does for Europe/Dublin, summer has dst() == 0 while winter has dst() == -1h. This condition therefore treats winter
    as the daylight side, so is_dst=True selects UTC+0 and False selects UTC+1, reversing the intended behavior and shifting ambiguous charts by an hour depending on the host's TZif
    encoding. Route negative DST through the offset comparison rather than treating every nonzero dst() as daylight.

  - [P3] Do not announce a clamp when substituting Porphyry — /Users/giacomo/dev/kerykeion/kerykeion/ephemeris_backend.py:364-366
    For polar Placidus/Koch calls, check_and_adjust_polar_latitude has already logged Latitude capped at 66° before this branch retries Porphyry at the original lat. The resulting logs
    contradict both the fallback record and the following warning that the real latitude was preserved; defer the logging clamp helper to the clamp_latitude branch or compute the gate
    without its side effect.

g-battaglia:
Both findings confirmed by measurement. The first one is more serious than it reads, and it means my previous commit was itself the regression.

## [P1] Confirmed — and it reverses my last fix

The two tz databases available on this machine genuinely disagree, exactly as described:

```
Europe/Dublin, 2026-10-25 01:30 (the fold)
  system database   IST dst=+1:00   GMT dst=0        -> "non-zero dst" picks IST  ✓
  packaged tzdata   IST dst= 0      GMT dst=-1:00    -> "non-zero dst" picks GMT  ✗
```

So `is_dst=True` returned summer on one host and **winter** on another. And the packaged database is precisely what a machine *without* a system one falls back to — the case the hard dependency exists for. A host-dependent one-hour error in a populated zone that has an ambiguous hour every year is much worse than the case I was trying to fix.

Before reverting I measured whether consulting the flag could ever have been necessary. Across 23 zones × every transition 1902–2037, on **both** encodings:

> the side carrying a **positive** `dst()` is always the side with the larger offset — **zero exceptions**

So the flag could never decide anything the offsets do not already decide, and the offsets carry it without the encoding. Reverted to the clock rule, with that reasoning now recorded in the docstring so it is not re-litigated.

**What this gives up** is the Kyiv 1941 case from the previous review. I no longer think it was a defect: that wall time repeats because the STANDARD offset changed, not because summer time ended, and modern tz data flags neither reading as daylight saving. There is no "is DST in effect" answer to give — the previous library answered from its own different model of that transition. It is now pinned as deterministic and documented as a case the parameter does not really apply to, rather than papered over with a rule that breaks Ireland.

## [P3] Confirmed and fixed

The gate was calling the clamp helper merely to *ask* whether a latitude sits inside the circle, and that helper logs while answering. Verified both paths after the fix:

```
lat 78, Placidus  ->  log contains "capped": False   record: substitute_system P->O, lat 78 used 78
lat 78, Gauquelin ->  log contains "capped": True    record: clamp_latitude,        lat 78 used 66
```

The gate now compares against the limit directly; the helper is called only where a latitude really is capped. Both branches pinned by a regression test.

---

Full suite: **9 992 passed, 1 387 skipped, 0 failed**, `ruff check` clean.

Worth saying plainly: this round caught a defect that the previous round *introduced*, and it was introduced because I acted on a review finding without first measuring whether the proposed remedy held everywhere. The measurement takes a minute and would have prevented it.

g-battaglia:
Codex Review (3)

• The calculations largely work, but polar charts are mislabeled as using the requested house system in several public output paths even when their cusps were calculated with Porphyry.

  Review comment:

  - [P2] Expose the substituted house system to output consumers — /Users/giacomo/dev/kerykeion/kerykeion/astrological_subject_factory.py:2414-2415
    When a polar Placidus/Koch chart reaches this block, cusps came from Porphyry, but only the fallback list is updated; houses_system_identifier and houses_system_name remain P/Placidus.
    ChartDrawer._get_domification_string, ReportGenerator, and astrological_subject_to_context still read those requested fields and do not render the fallback, so generated SVG, report,
    and XML outputs explicitly mislabel Porphyry cusps as Placidus—the updated Antarctic SVG fixture still says Domification: Placidus. Expose an effective system or make these consumers
    surface the main fallback.

g-battaglia:
Confirmed, and it is the most consequential finding of the three rounds — it undercut the justification for the design choice itself.

## Verified

```
polar Placidus chart, lat 78.2232
  houses_system_identifier : P
  houses_system_name       : Placidus
  polar_house_fallbacks    : substitute_system, P -> O
  first_house == ascendant : True        (the cusps really are Porphyry)
```

And the committed fixture, exactly as reported:

```
Arctic Subject - Natal Chart.svg      Domification: Placidus
Antarctic Subject - Natal Chart.svg   Domification: Placidus
```

The argument for substituting the house system was that the degradation is *declared* rather than silent. But `polar_house_fallbacks` is structured data no renderer reads — the SVG legend, the report table and the serialized context all take the system from those two fields. So the declaration existed only where nobody looks, and every rendered artifact stated something false. That is the same silence the fallback exists to end, moved one layer out.

## Fixed at the source, not per consumer

`houses_system_identifier` answers "which division are these houses", so after a substitution the honest answer is the substitute. The request is not lost — it was already on the fallback record, which is where a consumer that cares should look. Fixing it there rather than in each renderer also covers consumers neither of us enumerated.

```
polar, Placidus requested  ->  identifier O, name Porphyry   (record keeps P / Placidus)
polar, Whole Sign          ->  identifier W, unchanged, no fallback
lat 45, Placidus           ->  identifier P, unchanged, no fallback
```

Applied to the relocated-chart path too, where relocating *into* the polar circle is the commonest way a real chart meets this.

Both polar baselines regenerated: **three lines each, the legend only, no numeric drift**.

## The part worth keeping

This survived a full adversarial review pass and a golden file, and the reason is worth recording: `compare_chart_svg` matches numeric tokens with a tolerance and **does not look at text at all**. A legend naming the wrong house system passes it unchanged. The baseline could never have caught this, and regenerating it would have silently blessed the wrong label.

So the new coverage asserts the rendered string directly rather than trusting the fixture — `"Domification: Porphyry" in svg` and `"Domification: Placidus" not in svg` — alongside the subject-level assertions.

---

Full suite: **9 997 passed, 1 387 skipped, 0 failed**, `ruff check` clean.

g-battaglia:
Codex Review (4)

• The patch can shift current-time charts by one hour during negative-DST folds and can unintentionally make a location-specific polar house fallback persist into later non-polar chart
  calculations.

  Full review comments:

  - [P2] Preserve the exact instant in current-time folds — /Users/giacomo/dev/kerykeion/kerykeion/utilities.py:922-923
    On hosts using packaged tzdata's negative-DST encoding—for example Europe/Dublin, where the repeated hour has summer dst()==0 and winter dst()==-1h—from_current_time() computes
    bool(now.dst()) and passes it back through from_birth_data(). This resolver interprets that boolean as larger/smaller offset, so both sides are swapped and a chart created during the
    fall-back hour is one hour off; pass the already-resolved UTC offset through as from_iso_utc_time() does instead of re-deriving it from dst().

  - [P2] Keep polar substitutions from becoming permanent settings — /Users/giacomo/dev/kerykeion/kerykeion/astrological_subject_factory.py:2427-2429
    When a user requests Placidus or Koch at a polar latitude, overwriting this field with O changes more than the rendered label: RelocatedChartFactory and PlanetaryReturnFactory feed
    subject.houses_system_identifier back into later house calculations. A polar Placidus subject relocated or returned at a non-polar location therefore gets Porphyry cusps and loses its
    fallback record even though Placidus is valid there; retain or recover the requested identifier for recasts while exposing the used identifier for rendering.

g-battaglia:
Both confirmed, and both were introduced by my own previous two commits.

## [P2] The relabelling poisoned the recast path

Reproduced exactly as described:

```
polar Placidus subject     ->  identifier O   (after the previous commit)
relocated to Rome          ->  identifier O, name Porphyry, polar_house_fallbacks []
```

Rome casts Placidus perfectly well. So a substitution that belonged to one latitude followed the subject to a place where it had no reason to apply — and with an empty fallback list, nothing recorded why. Corrupting the numbers is worse than mislabelling them, and my previous commit traded the second for the first.

The field serves two consumers with opposite needs, which is the actual defect. It now answers only one:

- `houses_system_identifier` keeps the **request** — what a recast must inherit
- `effective_houses_system_identifier` / `_name` derive what the cusps came from — what rendering reads

```
polar, Placidus requested  ->  requested P/Placidus, effective O/Porphyry, SVG says Porphyry
relocated to Rome          ->  requested P, effective P, no fallback record
lat 45                     ->  the two views agree
```

The three known renderers (SVG legend, report table, serialized context) now read the effective view.

## [P2] `bool(now.dst())` was the same trap, one layer up

This is the portability problem removed from the resolver one commit ago, still live at the caller. Measured on Ireland's fold, reconstructing the instant the way `from_current_time` did:

```
system database    now=01:30 +0000  dst=0       -> bool=False  -> rebuilt 01:30 +0000   OK
packaged tzdata    now=01:30 +0000  dst=-1:00   -> bool=True   -> rebuilt 01:30 +0100   one hour off
```

The conversion has already resolved the offset unambiguously, so it is now passed straight down — the mechanism `from_iso_utc_time` has documented all along, and which `from_current_time` simply never adopted.

Pinned twice: once on the resulting instant under both databases, and once on the mechanism itself. The instant-only test would pass for every hour of the year but one, so on its own it would have been close to worthless.

## A note on the arc

Four rounds, six real defects, and **three of them were consequences of my own fixes** — the half-corrected subtraction, the host-dependent `dst()` rule, and now the recast poisoning. Each time the pattern was the same: I applied a remedy that was right about the case in front of me without measuring whether it held everywhere else.

One small thing that went right, though: the assertion I added last round pinning the relabelling is what failed this round and flagged the reversal. Pinning a decision makes changing it loud, including when the decision was mine and wrong.

---

Full suite: **10 002 passed, 1 387 skipped, 0 failed**, `ruff check` clean. No baseline moved this round.

g-battaglia:
Codex Review (5)

• The patch can select an ancillary Gauquelin fallback as the main house system and loses or misreports substitutions in midpoint composites and several presentation paths. These issues
  produce incorrect chart metadata and can mix different effective house systems silently.

  Full review comments:

  - [P2] Filter Gauquelin records out of the effective house lookup — /Users/giacomo/dev/kerykeion/kerykeion/schemas/kr_models.py:1122-1124
    When calculate_gauquelin=True at a polar latitude and the main system works there (for example Whole Sign), the only fallback is the ancillary G clamp. Because that record also lists
    "house_cusps", this loop returns G/Gauquelin sectors as the main chart's effective system, causing the context serializer and reports to mislabel Whole Sign cusps. Match the fallback's
    requested identifier to self.houses_system_identifier, or distinguish Gauquelin products explicitly.

  - [P2] Preserve effective systems when building midpoint composites — /Users/giacomo/dev/kerykeion/kerykeion/schemas/kr_models.py:1070-1073
    CompositeSubjectFactory still validates and copies only the requested house fields, then constructs midpoint models from self.__dict__, so this new inherited field defaults to an empty
    list. A polar Placidus subject can therefore be combined with a temperate Placidus subject even though their cusps are effectively Porphyry and Placidus, and two polar parents produce a
    composite that incorrectly reports effective Placidus while losing both fallback records. Compare and propagate effective systems or fallback provenance when constructing midpoint
    composites.

  - [P2] Use effective house labels in compact chart renderers — /Users/giacomo/dev/kerykeion/kerykeion/charts/chart_drawer.py:4253-4255
    This updates only _get_domification_string, while Synastry, Composite, and SingleReturn renderers use InfoSectionBuilder.build_houses_system_info, which still reads
    houses_system_identifier and houses_system_name. Consequently, a polar Placidus return or synastry chart renders Placidus Houses over Porphyry cusps; update the shared compact formatter
    and account for separate effective systems on dual wheels.

  - [P2] Use the effective system in the houses table title — /Users/giacomo/dev/kerykeion/kerykeion/report.py:573-575
    For a polar Placidus subject, the settings table now says Porphyry, but _houses_report still builds its title from houses_system_name, yielding a single report containing both Houses
    System | Porphyry and Houses (Placidus). Use the effective house-system name for the houses table title as well.

g-battaglia:
All four confirmed. Every one is a consequence of the effective/requested split I introduced last commit — the split was made at the model and then trusted to propagate, and it does not.

## The Gauquelin record was hijacking the lookup

```
polar Whole Sign + calculate_gauquelin=True
  requested  W
  EFFECTIVE  G  (Gauquelin sectors)      <- its own cusps never degraded
  records    [clamp_latitude, requested G, used G, affects ['house_cusps','angles']]
```

Both fallback strategies list `house_cusps` in `affects`, so my filter could not tell the chart's own record from the ancillary sector-ring one. It now matches on the **requested** identifier, which is precisely what separates them — the ancillary record asks for `G`, the main one asks for whatever the caller asked for.

```
Whole Sign + Gauquelin, polar  ->  effective W   (1 record, correctly ignored)
Placidus   + Gauquelin, polar  ->  effective O   (2 records, correct one chosen)
```

## Three consumers still read the requested fields

Confirmed all three, including one that was a defect **inside my own previous fix**: `_get_domification_string` had its translation key switched to effective while the displayed fallback name stayed on the requested field. The compact formatter behind synastry, composite and return wheels was untouched, and the report produced a document containing both `Houses System | Porphyry` and `Houses (Placidus)`.

After the fix the report is internally consistent, verified end to end.

## Composites validated the request and stopped

```
polar parent      requested P, effective O
temperate parent  requested P, effective P
composite         -> P, no fallback record        (both halves silently merged)
```

The factory already raises `Both subjects must have the same houses system` for mismatched requests. That refusal now applies to what the houses actually are, since a midpoint of Porphyry cusps and Placidus cusps is a composite in neither. Two substituted parents compose fine and the result states the substitute.

## On the pattern

Five rounds now, and the last three findings have all been downstream of my own previous fix rather than of the original change. That is worth naming: each fix was correct about the case in front of it and incomplete about the surface it touched. Introducing a distinction at the model does not propagate it — every consumer is a separate decision, and the ones that render are exactly the ones a reader believes.

---

Full suite: **10 007 passed, 1 387 skipped, 0 failed**, `ruff check` clean. No baseline moved.

PR #237, opened 2026-08-05
fix(sun-times): solar noon is the meridian transit, and now it is anchored

g-battaglia:
## The defect

`solar_noon` was the midpoint between sunrise and sunset. That equals the meridian **only while the Sun's declination is stationary** — so the old value was exactly right at the solstices and on the equator, where anyone spot-checking it would look, and wrong everywhere else.

Measured against two national observatories:

| | midpoint error |
|---|--:|
| Rome, equinox | +21.5 s |
| Ushuaia, equinox | +33.5 s |
| Reykjavík, April | **+62.4 s** |
| Rome, solstice | −0.1 s |
| Quito, equinox | −0.1 s |

And when the rise/set pair straddles local midnight the midpoint lands on the **wrong civil day** entirely — Singapore did.

Nothing in the library computes on this field (it is printed in the report, serialised into the AI context, and returned to callers), so the correction is visible rather than structural.

**New**: solar noon is now reported on **polar days and nights** too. A transit is a meridian crossing, not a horizon crossing — the Sun culminates on a day it never rises. `day_length` stays `None` there.

## Why it survived

Nothing tested it against anything outside this repository. Golden snapshots prove *constancy*, which a wrong number satisfies perfectly. So the fix ships with the two guards that were missing.

### Anchors — `tests/core/test_sun_times_anchors.py`

Values transcribed by hand from the **US Naval Observatory** and **IMCCE (Observatoire de Paris)**, same UTC civil day, capture date recorded, **no script can rewrite them**.

The tolerances are shaped by what the sources actually do, not by what would be convenient:

- they **agree** on sunrise and transit → held to 45 s of USNO
- they **disagree by ~2 min on sunset** (a horizon convention; and at Wellington it is USNO that is the outlier while IMCCE agrees with us) → every event must additionally lie inside the span the two of them bracket
- **above 60° latitude, no time-domain claim at all.** The Sun grazes the horizon, `dh/dt` collapses, and clock time stops being a well-conditioned way to state an error. That cut-off is *asserted*, not assumed: the same two services that agree to a minute at mid-latitudes differ by **10 and 11 minutes at Tromsø**, and `test_the_cut_off_latitude_is_earned` says so.

17 anchored sites, both hemispheres, four seasons, single-timezone extremes (Ürümqi), date-line side (Wellington), UTC-day straddles (Singapore, Tokyo).

### Altitude invariant — `tests/core/test_sun_times_altitude_invariant.py`

Never compares times. Takes the instant we return and asks **Skyfield** where the Sun was. (Honesty note from review round 3: the backend is itself built on Skyfield's machinery, so this is a second *use*, not a second implementation — what the suite isolates is everything kerykeion adds on top: bookkeeping, search, pairing, transit-vs-midpoint. Common-mode truth is the anchors' job.)

- true **upper limb at −33.59′**, spread **2.5″** across ten sites and four seasons
- **solar noon's hour angle under 0.1 s** of time
- **`is_diurnal` flips within a second** of the geometric centre crossing zero

Reverting solar noon to the midpoint kills **10** of these and **3** anchors, so neither file is decorative.

The documented gap between sunrise and diurnality (3.3 min at the equator, 4.4 at Rome, 8.2 at Reykjavík) is pinned there too, because the README now quotes it and prose drifts.

## Dependency: libephemeris 3.0.0rc15 → 3.0.0

Validated rather than assumed — see the second commit. Full cross-engine parity against a **file-backed** reference: identical verdicts, identical counts, 5213 quantities. Direct value-by-value diff: **16337 compared, 16267 bit-identical**. House cusps and angles did not move at all.

The 70 that moved are sidereal ayanamsas at 0.0014″ and the apsis family at 0.064″ — plus the eight Uranian points, which go from a disordered ±20″ scatter against the reference to a uniform +2..3″ bias (8 of 9 improve, Kronos by 20″). The 29 regenerated report fixtures are that and nothing else: **exactly eleven points changed position, all analytically-modelled bodies, none a planet, an angle or a cusp.**

⚠️ Worth recording: the parity grid does **not** cover the Uranian family, so a 36″ change passed 5213 comparisons unseen and only the report snapshots caught it.

## Documentation

The horizon convention is now stated where callers meet it — apparent upper limb, semidiameter from the real distance rather than a fixed 16′, standard-atmosphere refraction, level sea horizon at any elevation. Both the README and the model docstrings say plainly that sunrise and `is_diurnal` answer different questions and must never be derived from one another.

## Verification

- `poe test:extended` — **10684 passed**, 1390 skipped (10611 before, +73 new)
- `poe lint` — All checks passed
- `poe typecheck` — 0 errors, 1 pre-existing warning
- `poe analyze` — no issues in 108 source files
- mutation: midpoint reverted → 10 invariant tests + 3 anchors go red

https://claude.ai/code/session_011FVi6KXwFmp564FrQFbYRQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Solar noon now reflects the Sun’s actual meridian transit rather than a sunrise/sunset midpoint.
  - Solar-noon times remain available during polar day and night, while day length correctly remains unavailable when sunrise or sunset is absent.
  - Sunrise and sunset calculations now consistently apply upper-limb and atmospheric-refraction conventions.

- **Documentation**
  - Clarified solar-noon semantics, polar-day behavior, sunrise/sunset calculations, and the distinction between sunrise and diurnal status.

- **Tests**
  - Added broad validation against astronomical reference observations and independent solar-geometry checks.

---

## Round 3 (blind adversarial review, all findings fixed in `2ca86c6b`)

- The rewritten `_polar_state` docstring still carried a name this project never uses, in a line this branch itself added — rewritten to point at the constant whose note owns the 0.006° gap.
- The altitude suite's "shares no code" independence claim was **false** (see note above); its module docstring now states exactly what it does and does not isolate.
- **A transit on the wrong civil day was invisible to the whole suite**: every comparison unwraps to the nearest day. Proven by reverting solar noon to the midpoint — Singapore answered 2026-08-06 for a 2026-08-05 request and stayed green. A new date-pinning test covers all 19 grid cases; the same mutant now fails 5 of them.
- The module-level transit blindfold in the mocked suite also covered the two **real-ephemeris** classes, silently zeroing their transit coverage — the round-2 commit message's "three tests still call the backend for all six helpers" was wrong (five real, one fake). Both classes now shadow the fixture with a no-op.
- The declination count is **19, not 17**: the dual-return fixture carries Mean Lilith and Mean Priapus in both its tables — the exact duplicate-cell trap the CHANGELOG paragraph warns about caught its own correction.
- The factory docstring recovered the transition-day qualifier ("day_length can exceed 24 h") it had dropped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,7 +95,11 @@
   Descendant 15, Imum Coeli 15) across 28 fixtures, plus **122 non-angle speed
   cells** — the Uranian points and, not previously named anywhere,
   True Lilith 12, True Priapus 12 and Interpolated Perigee 11 — and
-  **17 declination cells**. `ascmc_speed` comes from `houses_ex2`; the parity
+  **19 declination cells** (Mean Lilith 8, Mean Priapus 8, Hades 1, White
+  Moon 1, Admetos 1 — a first count said 17, having taken the dual-return
+  fixture's Lilith and Priapus rows once when that file carries them in both
+  its tables: the very duplicate-cell trap this paragraph warns about).
+  `ascmc_speed` comes from `houses_ex2`; the parity
   front excludes cusp and angle speed by design, so the campaign could not have
   seen any of it.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,17 +52,36 @@
 
 ### Changed
 
-- `libephemeris` pinned to **3.0.0** (from `3.0.0rc15`). Validated rather than
-  assumed: a full cross-engine parity campaign against a file-backed reference
-  returned identical verdicts and identical counts over 5213 compared quantities,
-  and a direct value-by-value diff of 16337 quantities found 16267 bit-identical.
-  House cusps and angles did not move at all. The 70 that did are sidereal
-  ayanamsas at 0.0014″ and the apsis family at 0.064″, and the eight Uranian
-  points, which go from a disordered ±20″ scatter against the reference to a
-  uniform +2..3″ bias — 8 of 9 improve, Kronos by 20″. The report fixtures were
-  regenerated for that reason and no other: exactly eleven points changed
-  position, all of them analytically-modelled bodies, none of them a planet, an
-  angle or a cusp.
+- `libephemeris` floor raised to **>=3.0.0,<4** (from the exact `==3.0.0rc15`).
+  Validated rather than assumed, and the evidence is worth stating precisely
+  because a first draft of this entry overstated it.
+
+  The cross-engine parity campaign, run against a file-backed reference before
+  and after, returned **identical per-domain counts over 5213 compared
+  quantities** and no divergence appearing or disappearing. Note the verdict it
+  returns is RED in both runs — the pre-existing sidereal-ayanamsa and
+  deep-time offenders — so "identical" means unchanged, not clean. The campaign
+  was also run under kerykeion `6.0.0a75`, the version its lockfile pins.
+
+  Two further measurements come from ad-hoc scripts rather than from that
+  campaign, and are reproducible but not archived in any repository: a
+  value-by-value diff of 16337 quantities found 16267 bit-identical, and the
+  eight Uranian points move from a ±20″ scatter against the reference to a
+  uniform +2..3″ bias (8 of 9 improve, Kronos by 20″). The parity grid does not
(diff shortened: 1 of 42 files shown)


g-battaglia:
@coderabbitai review

PR #239, opened 2026-08-07
feat(charts): modern declutter is measured, optimal, and content-aware

g-battaglia:
## What

Three upgrades to the modern wheel's collision handling, each derived from **browser measurements** instead of guessed constants:

### 1. Separations are measured
`PLANET_MIN_SEPARATION` 8.0 → **7.25**, and the dual rings' shared hardcoded `10.0` becomes `SYN_OUTER_MIN_SEPARATION` **5.75** / `SYN_INNER_MIN_SEPARATION` **7.5** — the values `scripts/measure_modern_separation.py` proves safe by rendering adversarial worst-content clusters and reading real ink boxes out of Chrome.

### 2. Placement is displacement-optimal (PAVA)
The forward-walk resolver anchored the first planet of a cramped run and pushed everyone else ahead. `_resolve_planet_collisions` is now an **isotonic regression** (pool-adjacent-violators on separation-deflated coordinates): the provable least-squares placement preserving zodiacal order and pairwise gaps. Cramped runs spread symmetrically around their true center of mass; a wraparound guard clips the fit when a layout would close the cut gap.

| fixture | worst displacement | mean |
|---|---|---|
| natal stellium 2000-02-26 | 7.4° → **4.6°** | 2.0° → **1.2°** |
| natal all-points (52 pts) | 30.0° → **15.8°** | 8.0° → **4.9°** |
| synastry outer ring | 43.4° → **20.7°** | 15.3° → **7.4°** |

### 3. Pairs reserve what they draw
Each adjacent pair's separation comes from its own measured ink — glyph artwork, the exact degree/minute strings, an `rx` row only when retrograde — capped at the ring's measured ceiling so no layout can regress. Both ink axes count: clusters stay upright while the wheel turns, so at the worst orientation neighbours pinch on the **diagonal**, where width-only spacing under-reserves by up to 40% (caught by the adversarial harness before shipping, twice — see below).

## How it was validated (browser-first)

- `scripts/measure_modern_separation.py` gained three modes: `floor` (how tight can a constant go before ink touches), `adversarial` (render mixed narrow/wide, retro/direct clusters with the real drawing code and pixel-measure every gap), `dump-ink` (rasterize all 56 glyphs + 12 signs + 91 row strings and write `kerykeion/charts/glyph_ink_metrics.py`; `poe regenerate:glyph-ink`).
- The adversarial probe **failed two intermediate models** — width-only spacing, then high-resolution-only text metrics (font hinting at wheel sizes) — and shipped only after every engineered worst case keeps ≥ 0.24 wheel-units of daylight.
- `scripts/report_modern_displacement.py` prints the before/after table above.
- `tests/core/test_modern_decluttering.py`: 57 tests, including block-centering (the PAVA signature), an independent O(n²) PAVA cross-check, perturbation/KKT optimality, displacement-regression pins, and content-aware invariants.

## Regeneration

Modern SVG baselines (62), `tests/data/v6_gallery`, `docs/charts` — non-modern files untouched (their pre-existing staleness is left for a separate cleanup).

## Checks

- `poe test:core`: 5518 passed · `test_chart_parametrized`: 76 passed
- `poe lint` / `poe typecheck`: clean

https://claude.ai/code/session_018cyj31AKfuV526TmRa63o2


g-battaglia:
Second round, prompted by review: the spacing was still wider than the ink justified.

**Orientation-aware separations.** The per-pair requirement was the all-orientation worst case (`hypot(widths, heights)`), but every rendered pair sits at a known orientation: near the top of the wheel neighbours sit side by side (widths bind), at the sides they stack vertically (only the much smaller text heights bind). Requirements are now evaluated where the pair actually sits, solved to a fixed point with a ratcheted refinement loop that only accepts a placement after re-evaluating it against itself. Side-of-wheel text pairs pack ~40% tighter.

**The wheel pins its text font.** The adversarial harness caught something better than a bug in the model: the modern wheel never declared a `font-family`, so the same chart rendered serif standalone and inherited whatever font the embedding page used — the probe page's own `monospace` was bleeding into the SVGs under test. No spacing table can be sound against an unknown font. The wheel root now declares `Arial, Helvetica, sans-serif` (metric-compatible with Liberation Sans on Linux) and the ink tables are measured in that stack. **Visible change:** standalone charts now render cluster/cusp text in sans-serif.

Updated displacement table (vs the branch's starting point):

| fixture | worst | mean |
|---|---|---|
| natal stellium | 7.4° → **2.7°** | 2.0° → **0.7°** |
| natal all-points | 30.0° → **14.3°** | 8.0° → **4.5°** |
| synastry outer ring | 43.4° → **13.0°** | 15.3° → **3.9°** |

Adversarial probe: PASS on 192 synthetic wheels across eight orientations, no ink touching. 5518 + 76 tests green, lint/typecheck clean, modern baselines regenerated.

g-battaglia:
Review round (xhigh, 32 agents): 15 verified findings, all fixed in 4ec91835.

**The theme:** the font pin had reached the production wheel but not the measurement instruments. The probe page's `monospace` was still bleeding into the SVGs under test, the dump rasterizer measured the six text-based glyphs (As/Mc/Ds/Ic/Av/Vx) in default serif, and the full-chart template's 28 text nodes had no pin at all. All instruments now measure under the pinned stack; floors re-measured honestly: ink touches at **6.25 / 5.00 / 6.25°** — the shipped ceilings keep 0.75–1.25° of slack, documented as headroom for wider-inking fallback fonts (Liberation Sans is now named in the stack, unquoted because the SVG post-processing rewrites quote characters).

**Also fixed:**
- Code-injection sink in the `--dump-ink` POST endpoint (payload now validated against the script's own glyph/text catalog before touching generated Python).
- Adversarial gate now fails below 0.2 wheel-units of daylight (was: any non-negative gap); probe pads stroke-only glyph boxes by their stroke reach. Healthy state measures 0.32+ everywhere.
- Orientation refinement that exhausts its rounds logs a warning instead of shipping an unvalidated placement silently.
- `SIGN_INK` lookups get the same conservative fallback as glyphs and texts.
- Consolidation: `svg_metadata.py` is the one kr:-attribute parser (was 3 regex copies); `_pair_required_separation` is shared by resolver and tests; natal row Ys are single-source module constants pinned by the live copy; straight-tether threshold shared; generated-module docstring corrected; CHANGELOG entries merged with final figures.

5522 + 76 tests green, lint/typecheck clean, modern baselines regenerated (font attribute + Main_Chart pin), XML validity spot-checked.

PR #247, opened 2026-08-11
refactor: one package per domain, plus a visible v5 upgrade path

g-battaglia:
## What and why

`kerykeion/` was organised in two incompatible ways at once: 33 domains lived in packages with a re-exporting `__init__.py`, while 16 modules — often the largest in the library — sat as single files in the package root. The naming inside the packages was split too: `firdaria/factory.py` next to `eclipses/eclipse_factory.py`.

After this PR, **`kerykeion/__init__.py` is the only `.py` file in the package root**. Everything else lives in a domain package with a uniform internal naming rule.

Public behaviour is unchanged.

## The naming rule

1. One file at the top level: `kerykeion/__init__.py`.
2. The package is the domain, the file is the role: `X_factory.py` → `X/factory.py`.
3. Inside a package, no prefix repeating the package: `eclipses/eclipse_factory.py` → `eclipses/factory.py`.
4. `factory.py` is reserved for the factory the package is named after; a second technique keeps its own domain name (`secondary_progressions/solar_arc.py`).
5. Nothing else is renamed. A name that does not violate 2 or 3 and carries no legacy prefix stays as it is.

Rule 5 is why `orb_utils.py`, `config_constants.py`, `chart_defaults.py`, `rulers.py`, `catalog.py`, `nakshatra_*.py` and `draw_*.py` are untouched: their prefix is a real distinction, not a repetition.

## What moved

The 16 root modules became packages — `astrological_subject/factory.py`, `report/generator.py`, `geonames/fetcher.py`, `utilities/core.py`, `ephemeris_backend/backend.py`, and so on — plus 23 renames inside existing packages, and `schemas/` dropping the legacy `kr_` prefix (`kr_models` → `models`, `kr_literals` → `literals`, `kerykeion_exception` → `exceptions`).

`kr_types/`, the v4.x shim deprecated since v5, is removed. `setup.cfg` goes with it: it configured flake8, which is not among the project gates, and named a file that no longer exists.

Two names keep their full form on purpose. `swisseph_setup/` stays, because `python -m kerykeion.swisseph_setup` is documented — the CLI now reaches it through `__main__.py`, with an explicit `prog=` so `--help` prints what it printed before. `ephemeris_backend/` stays too: `ephemeris/` would have read as a sibling of `ephemeris_data/`, moving 142 references for a name that is less clear, not more.

## The public API does not move

`kerykeion.__all__` exports the same 120 names, and a snapshot of every export plus the JSON schema of every Pydantic model among them is **byte-identical** before and after. That is the load-bearing check for the Astrologer API, whose OpenAPI document is built from these models.

`AstrologicalSubjectFactory` moved file, but `from kerykeion import AstrologicalSubjectFactory` is untouched — and so is `from kerykeion.utilities import localize_naive`, thanks to a flat facade on that package.

## The new gate

`scripts/check_import_graph.py`, wired into `poe check`, covers what mypy and pyright structurally cannot: they only look at `kerykeion/`, so the ~1500 module paths in `tests/`, `scripts/` and `examples/` were unchecked, and the ones written **inside strings** were invisible to every existing tool.

Four modes, each aimed at a failure that otherwise stays silent:

- **`--paths`** — every `kerykeion.*` path resolves, whether written as an import or as a string constant.
- **`--patch-targets`** — no `mock.patch` target sets an attribute on a *package*. When `foo.py` becomes `foo/`, patching `"kerykeion.foo.Thing"` still succeeds while the code under test reads `kerykeion.foo.impl.Thing`: the patch quietly does nothing.
- **`--loggers`** — every `logger=` string in the tests names a logger some module actually registers. `caplog.at_level(logger="pkg")` keeps capturing through the hierarchy after a move, so a stale name leaves a filter that verifies nothing.
- **`--fresh-imports`** — each module imports on its own in a cold interpreter. pytest, mypy and `pkgutil.walk_packages` always import `kerykeion/__init__.py` first, so none of them can see a cycle reachable only leaf-first.

It earned its keep three times: 7 `caplog` filters left pointing at a package, 5 tests patching through a module alias, and — before any file moved — an `ImportError`-based `hasattr` in `kerykeion/__init__.py`.

## Fixes found on the way

- `test_no_warning_when_se1_present` asserted a string was **absent** from stderr without checking `returncode`, so it passed even when the subprocess died of `ImportError` — blind exactly where it mattered.
- `mypy_path = "kerykeion"` also resolved `utilities`/`report`/`schemas` as top-level modules, which would have masked any import written without the package prefix.
- ruff `INP001` on `kerykeion/`: a forgotten `__init__.py` silently yields a working PEP 420 namespace package, and nothing else catches it.
- The composite polar snippet called `CompositeSubjectFactory(polar_a, polar_b)` with neither name ever bound. Pre-existing — both were already undefined before this branch.
- `SIGN_CODES` is now exported from `kerykeion.schemas`, so the one consumer that needs it no longer imports a private module path.

## Verification

| Gate | Result |
|---|---|
| `poe test:core` | 5662 passed, 144 skipped |
| `poe test:extended` | 10888 passed, 1390 skipped |
| `poe lint` / `poe analyze` | clean |
| `poe imports` / `imports:fresh` | clean (1517 paths, 51 patch targets, 130 cold imports) |
| `poe docs:check` | 120/120 exports documented |
| `poe docs:snippets` | 400/400 |
| `poe build:smoke` | wheel renders a chart — templates and themes ship correctly |
| public API + Pydantic schema snapshot | byte-identical |

`poe typecheck` reports one pyright error in `report/generator.py`. It is **pre-existing**: the same error appears at `7e3f7733` with the code byte-identical, and this branch also removes a warning that was there.

---

# Part two: making the silent half of the v5 upgrade visible

The layout change above is invisible to anyone importing `from kerykeion import X`. While verifying that, a separate and worse problem surfaced, so it is fixed here too.

Upgrading from v5 breaks in two ways, and only one is loud. The loud one is the removed API — `AstrologicalSubject` and friends raise an `ImportError` naming their replacement, and that case takes care of itself. The quiet one hits the users who **had already migrated correctly**: v6 changed defaults that alter results, so code written against the v5 factory API keeps running and returns different numbers.

| What | v5 | v6 | Effect on unchanged v5 code |
|:-----|:---|:---|:----------------------------|
| Default active points | 18 | 14 | `Descendant`, `Imum_Coeli`, `True_South_Lunar_Node`, `Mean_Lilith` are no longer computed |
| Sun/Moon aspect orbs | no adjustment | `+1.5°` | Aspects appear or disappear near the orb boundary |
| `ChartDrawer` style | `"classic"` | `"modern"` | A different drawing, and a different saved filename |

All three verified against `git show v5:…`, not inferred. Nothing raises, nothing warns: the user finds out weeks later, looking at a chart that does not match.

**1. The removed-name `ImportError` now names the changed defaults too.** It is the one moment we know the reader is looking, so it reports both halves of the upgrade instead of only the one they asked about.

**2. `kerykeion.settings.V5_DEFAULT_ACTIVE_POINTS`** restores the old point set explicitly, for anyone needing numerical continuity. Deliberately a frozen historical record, not a maintained preset — it will not follow future changes to the defaults, and the docstring says so. `DEFAULT_ACTIVE_POINTS`, `DEFAULT_ACTIVE_ASPECTS` and `ALL_ACTIVE_POINTS` are re-exported alongside it so callers depend on the package facade rather than on `config_constants`' filename.

**3. A "What changes in the results" section in the migration guide**, with runnable examples for all three — including `point_orb_adjustments={}` for v5-equivalent orbs and `style="classic"` for the v5 drawing. The guide previously covered renamed APIs only and said nothing about behaviour.

### No v5 compatibility wrapper, deliberately

`kerykeion/backword.py` still exists in `git show v5:` and could be restored cheaply, with its 1205-line test suite. Against it: with the defaults staying breaking, a wrapper removes the *signal* without removing the *change* — the code would run and the results would still differ. The `DeprecationWarning` meant to compensate is hidden by default in Python unless the call sits in `__main__`, so in any real application it is invisible. The `ImportError` is doing useful work: it is what sends people to the guide.

Worth recording: those classes were *already* deprecated wrappers in v5, and the layer was removed deliberately in `2309807c`. Restoring it means signing up to maintain it again, and to have the same discussion at v7.

The eight new tests pin the claims rather than the implementation: that v6 only ever *dropped* points from the default set (so the guide's "18 → 14, nothing added" cannot silently become false), that every v5 point is still computable today, and that passing the constant genuinely brings the four missing points back.

## Companion change

`v6.astrologer-api` is pinned to `kerykeion==6.0.0a83` from PyPI, so it is unaffected until the bump. Its imports have been aligned on branch `chore/kerykeion-package-layout`, using facades valid in **both** the pinned version and this tree — verified against each (1217 passed on a83; the suite also runs against a wheel built from this branch), with `openapi.json` regenerating byte-identical.

One note for whoever does the bump: `tests/baselines/advanced_report` will need regenerating, as it did for a81. That divergence does not come from this PR — the report grew Motion and OOB columns in `feat/report-enrichments`, already merged here but not yet published.


g-battaglia:
@coderabbitai review

PR #248, opened 2026-08-11
feat: make the silent half of the v5 upgrade visible

g-battaglia:
> Stacked on #247 — base is `refactor/package-layout`. GitHub will retarget this to `alpha/v6` once that merges.

## Why

Upgrading from v5 breaks in two ways, and only one of them is loud.

The loud one is the removed API: `AstrologicalSubject`, `KerykeionChartSVG`, `NatalAspects`, `SynastryAspects` raise an `ImportError` naming their replacement. That case takes care of itself — the user reads the error and fixes the call.

The quiet one hits the users who **had already migrated correctly**. v6 changed defaults that alter results, so code written against the v5 factory API keeps running and returns different numbers:

| What | v5 | v6 | Effect on unchanged v5 code |
|:-----|:---|:---|:----------------------------|
| Default active points | 18 | 14 | `Descendant`, `Imum_Coeli`, `True_South_Lunar_Node`, `Mean_Lilith` are no longer computed |
| Sun/Moon aspect orbs | no adjustment | `+1.5°` | Aspects appear or disappear near the orb boundary |
| `ChartDrawer` style | `"classic"` | `"modern"` | A different drawing, and a different saved filename |

Nothing raises, nothing warns. The user finds out weeks later, looking at a chart that does not match, with nothing connecting it to the upgrade.

All three are verified against `git show v5:…`, not inferred.

## What this does

**1. The `ImportError` now names the changed defaults too.** It is the one moment we know the reader is looking, so it reports both halves of the upgrade rather than only the one they asked about:

```
'AstrologicalSubject' was removed in v6. Use the factory instead:
    from kerykeion import AstrologicalSubjectFactory
    subject = AstrologicalSubjectFactory.from_birth_data(...)

Note: v6 also changed defaults that affect RESULTS, not just imports:
  - active points: 18 -> 14 (Descendant, Imum_Coeli, True_South_Lunar_Node,
    Mean_Lilith are no longer active unless requested)
  - Sun/Moon aspect orbs: +1.5 degrees by default (new in v6)
  - chart style: 'classic' -> 'modern'
Porting the call above does not restore v5 output. See 'What changes in the
results' in the guide; kerykeion.settings.V5_DEFAULT_ACTIVE_POINTS restores
the old point set.
Migration guide: https://www.kerykeion.net/content/docs/migration
```

**2. `kerykeion.settings.V5_DEFAULT_ACTIVE_POINTS`** restores the old point set explicitly, for anyone who needs numerical continuity. Deliberately a **frozen historical record, not a maintained preset**: it will not follow future changes to the defaults, and the docstring says so. `DEFAULT_ACTIVE_POINTS`, `DEFAULT_ACTIVE_ASPECTS` and `ALL_ACTIVE_POINTS` are re-exported from `kerykeion.settings` alongside it, so callers can depend on the package facade rather than on `config_constants`' filename.

**3. A "What changes in the results" section in the migration guide**, with runnable examples for all three — including `point_orb_adjustments={}` for v5-equivalent orbs and `style="classic"` for the v5 drawing. The guide previously covered renamed APIs only and said nothing about behaviour, which is the part that bites people who did everything right.

## What this deliberately does not do

**No v5 compatibility wrapper.** `kerykeion/backword.py` still exists in `git show v5:` and could be restored cheaply, along with its 1205-line test suite. I recommend against it, and the reasoning is worth recording:

With the defaults staying breaking, a wrapper removes the *signal* without removing the *change*. The code would run and the results would still differ. The `DeprecationWarning` meant to compensate is hidden by default in Python unless the call happens in `__main__` — so in any non-trivial application it is invisible.

The `ImportError` is doing useful work: it is what sends people to the guide, where they also learn the defaults moved. The quiet break is the one that needed the effort.

Worth noting too: those classes were *already* deprecated wrappers in v5, emitting `DeprecationWarning`, and the layer was removed deliberately in `2309807c`. Restoring it means signing up to maintain it again, and to have the same discussion at v7.

## Verification

| Gate | Result |
|---|---|
| `poe test:core` | 5662 passed, 144 skipped (8 new tests) |
| `poe docs:snippets` | 400/400 — the three new guide examples actually run |
| `poe docs:check` | 120/120 exports documented |
| `poe lint` / `poe analyze` / `poe imports` | clean |

The new tests pin the claims rather than the implementation: that the constant holds exactly the 18 v5 points, that v6 only ever *dropped* points from the set (so the guide's "18 → 14, nothing added" cannot silently become false), that every one of those points is still computable today, and that passing the constant genuinely brings the four missing points back.

https://claude.ai/code/session_01MQy5QpvBWe1Vk9LyHbBRnK

### DIFF
--- a/kerykeion/__init__.py
+++ b/kerykeion/__init__.py
@@ -341,6 +341,22 @@ __all__ = [
 # =============================================================================
 _MIGRATION_GUIDE_URL = "https://www.kerykeion.net/content/docs/migration"
 
+# Appended to every removed-name message below. Rewriting the call is only half
+# the upgrade: v6 also changed defaults that alter the numbers, and those change
+# silently for code that already used the v5 factories correctly. This error is
+# the one moment we know the reader is looking, so it says both things.
+_BEHAVIOUR_CHANGES_NOTE = (
+    "\n\n"
+    "Note: v6 also changed defaults that affect RESULTS, not just imports:\n"
+    "  - active points: 18 -> 14 (Descendant, Imum_Coeli, True_South_Lunar_Node,\n"
+    "    Mean_Lilith are no longer active unless requested)\n"
+    "  - Sun/Moon aspect orbs: +1.5 degrees by default (new in v6)\n"
+    "  - chart style: 'classic' -> 'modern'\n"
+    "Porting the call above does not restore v5 output. See 'What changes in the\n"
+    "results' in the guide; kerykeion.settings.V5_DEFAULT_ACTIVE_POINTS restores\n"
+    "the old point set."
+)
+
 _V5_REMOVED_NAMES = {
     "AstrologicalSubject": (
         "'AstrologicalSubject' was removed in v6. Use the factory instead:\n"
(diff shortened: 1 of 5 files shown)


g-battaglia:
Merged into #247 — everything now ships as a single PR, as requested. The commit is unchanged (b78c18da), just carried over onto refactor/package-layout.

PR #251, opened 2026-08-12
feat(charts): name the two stations, carry point state in the metadata, and add six opt-in marks

g-battaglia:
## What this adds

The models have known a great deal about each point since v6 — its speed, its declination, whether it is out of bounds, how it is moving relative to its mean motion — and every chart data model computes its angularities and stelliums. None of it reached the SVG. A consumer reading the markup had to re-fetch the JSON to say anything the wheel already knew, and a reader looking at the chart could not see any of it at all.

Three things, in that order of confidence:

**1. The two stations are now named.** `motion_state` gains `stationary_retrograde` and `stationary_direct`. The distinction matters and the sign of the speed alone cannot make it: both stations are approached from one side and left on the other, so what separates them is the *trend*. `classify_motion_state` takes an optional `speed_sampler` and consults it for a second sample a day out — but only for a body already inside the stationary band, so the extra ephemeris call is never spent on ordinary motion. Without a usable sample the generic `stationary` stands; absence of a claim, never a guess.

The band also became symmetric. It previously covered only the forward side, because a negative speed had already answered `retrograde` one branch earlier — so a planet creeping backwards at a hundredth of its mean motion was reported as plainly retrograde, hiding the very event the reader was looking for.

**2. Point state and chart analyses ride in the `kr:` metadata, always.** This is data, not decoration, so no option gates it: `kr:motionstate`, `kr:speed`, `kr:declination`, `kr:oob`, plus `kr:magnitude`/`kr:nearpoint`/`kr:orb` on fixed stars, and — injected onto the finished markup the way projected houses already are — `kr:angularity`, `kr:angularitydistance` and `kr:stellium`.

**3. Six opt-in marks, every one off by default.** Stations on the wheel, out-of-bounds badges in the point tables, dashes on separating aspect lines, the synastry score, the ayanamsa offset, and a note when a polar chart could not use the house system it was asked for.

## Design notes worth reviewing

**Off has to be genuinely off.** These add marks a reader did not ask for, and a chart that gains them on upgrade is a chart whose look changed under its owner. That is asserted directly rather than described: passing all six options their own defaults reproduces the previous SVG byte for byte, in both styles.

**The modern wheel is the standard, so it is not treated as an afterthought.** A station takes the marker row the retrograde label already occupies, labelled SR or SD, and recolours the cluster the way a retrograde does. It can share that row because the two states are mutually exclusive by construction, and both labels are two characters wide like RX — so the browser-measured separation model holds without remeasuring. The classic wheel puts the same SR/SD at the foot of the glyph, where its ℞ sits.

**The emitter lives beside the parser.** Three serializers draw celestial points, and an attribute that two of them emit is worse than one none of them emit: a consumer cannot tell a body that has no such state from a chart style that forgot to say so. `point_state_attributes` in `svg_metadata.py` is the one sentence all three speak. Values are rounded there too — raw floats would carry backend noise into every byte-compared baseline downstream for digits no reader uses.

**Attribute names are lowercase letters only**, and a test pins that across both styles and all three output templates. Consumers rewrite the namespace with a general pattern rather than an allow-list, so a name carrying an underscore would be dropped in silence rather than rejected loudly.

**Each mark is silent where it has no referent** — a chart with no station, synastry data that never computed a score, a tropical zodiac, a house system that was honoured. A mark that appears anyway is worse than one that never appears.

## The baseline diff

All 405 committed baselines changed, and they were already stale before this branch: the comparator's cross-backend tolerance treats any non-numeric difference as "structural, not a precision issue" and returns without comparing, which is what let a repo-wide change pass the whole suite untouched. That tolerance is deliberate and stays; this brings the committed artefacts back in line with what the code produces, so the next person to regenerate sees only their own diff.

Seventeen baselines are new, in two groups. Eleven isolate one mark each, in both styles — those answer "did this mark break". Six more are whole charts with every option on, which answer the different question of what someone who turns the feature on actually gets.

The same four subjects also went into the showcase galleries, because both of them demonstrated the marks by their absence: every committed chart was drawn with the options off, so a reader could read six descriptions and never see one. They are now embedded in the README and the charts page side by side in both styles, and the v6 gallery has its own section for them.

It takes four subjects rather than one because a mark draws nothing where there is nothing to mark. All six options are on for each, so none can claim something its own sky does not have: Mercury at its August 1990 station with Uranus past the obliquity shows the wheel marks and no ayanamsa; the sidereal chart shows an offset and no station; the Longyearbyen chart admits a substituted house system; the synastry prints a score.

## Verification

- `poe test:core` — 5773 passed, 144 skipped
- `poe lint` — clean; `poe analyze` (mypy) — no issues in 131 source files; `poe imports` — clean
- `poe typecheck` (pyright) reports one error in `report/generator.py:296`. **Pre-existing**: verified identical on `alpha/v6` in a scratch worktree, and untouched by this branch's diff to that file.
- **Cross-checked against the independent implementation already in this repo.** `RetrogradeStationFactory` locates the exact instant a planet's speed crosses zero and labels it SR or SD; `classify_motion_state` answers the same question from one chart's speed plus a single extra sample. They share no code, and they agree on all six Mercury stations of 1990 — including the April one, correctly named while the speed is still negative, which the previous branch order would have called a plain retrograde. That agreement is now a test.
- `poe docs:check` — 120/120 exports covered; `poe docs:snippets` — 408/408 executed (400 before this branch; the eight new ones are mine and all pass)
- Studio's own preprocessing regex simulated against the output: zero attributes dropped.
- 18 combinations of chart type × style × output template render as well-formed XML with every option on.

## Downstream

The extended `MotionState` literal is a breaking change for consumers that match exhaustively on its values. The API and Studio branches that consume this are prepared but gated on this release.

https://claude.ai/code/session_016CXv9wr5mmkAkzWiZDRKX9


g-battaglia:
@coderabbitai full review

Please scope this review to the `kerykeion/` directory only — the core library code. Ignore everything outside it.

Concretely, please review:

- `kerykeion/motion/state.py`
- `kerykeion/astrological_subject/factory.py`
- `kerykeion/charts/svg_metadata.py`
- `kerykeion/charts/drawer.py`
- `kerykeion/charts/draw_modern.py`
- `kerykeion/charts/draw_planets.py`
- `kerykeion/charts/utils.py`
- `kerykeion/charts/themes/*.css`
- `kerykeion/schemas/literals.py`, `kerykeion/schemas/models.py`, `kerykeion/schemas/settings_models.py`
- `kerykeion/settings/translation_strings.py`
- `kerykeion/report/generator.py`

And please skip: `tests/`, `docs/`, `site/`, `scripts/`, `examples/`, `README.md`, `CHANGELOG.md` and `kerykeion/llms.txt`.

The reason for the narrow scope: 405 of the 435 changed files are regenerated SVG baselines, and reviewing them would bury the roughly 1,500 lines of actual logic. Those baselines are machine-generated output, not hand-written code.

Three things I would especially value a second opinion on:

1. **`classify_motion_state` in `kerykeion/motion/state.py`.** The stationary band now brackets zero on both sides and is tested *before* the sign of the speed — a deliberate behavioural change. Inside the band an optional `speed_sampler` is consulted for a second sample a day out to tell the two stations apart. Is the fallback path (no sampler / sampler returns `None` / speed unchanged) airtight, and is the ordering of the branches right?

2. **The sampler closure in `kerykeion/astrological_subject/factory.py`** (in the motion-state enrichment block). It closes over `julian_day` and `_iflag` from the enclosing scope and is built per point inside a loop. Please check the closure captures what it should and that the extra `calc_ut` cannot fire outside the stationary band.

3. **`_inject_analysis_metadata` in `kerykeion/charts/drawer.py`.** It annotates finished markup by regex, keyed on `(kr:horoscope, kr:slug)`. Please check the dual-wheel case, where the same slug appears in both rings, and whether appending attributes after the `transform` can break any consumer that parses these tags.

PR #253, opened 2026-08-24
fix(review): twenty rounds of review, and the defects the fixes themselves introduced

g-battaglia:
> Replaces #252, closed because the automated review was reading it incrementally. Same branch, same commits: here CodeRabbit re-reads the whole diff from scratch, which is the point.

## What

It started as the fix for **13 defects** found by a code review of `alpha/v6` against `main`. It became a twenty-round review→fix→verify cycle that found many more — and most of the defects in the last rounds were **introduced by the fixes of the earlier rounds**, always in the same area. All are closed, each with a mutation proof: restore the defect and the test that pins it must fail, on its own.

Twenty-eight commits. **12,121 green tests**, `ruff` and `mypy` clean, 543 charts in the validation gallery.

---

## The thread that ties them

Almost every defect has the same shape: **a rule written in two places, and only one of the two updated.**

- `get_planet_house` had learned that cusps do not always ascend. The composite chart kept a **private copy** that had not learned it: on a descending ring it got **ten points out of ten** wrong, while another field of the same object — which used the shared function — gave the right answer.
- `normalize_degree` had been rewritten to close a trap (`% 360` answers *exactly* 360.0 for a tiny negative). The trap stayed open in **four other files**, one of them written during this PR.
- The list of themes × sidereal modes asked the test for ten combinations and produced **three** in the script: the seven without a file skipped with a message that reads in the logs like a chart that cannot be computed.
- The predicate "is this angle its own cusp?" took the index of the **angle** and folded it onto a cusp by itself. A new caller, with four angles to ask about, passed it **cusp** numbers: Imum Coeli and Descendant ended up measured against the tenth, said no every time, and never got the answer that predicate exists for.

The cure was not yet another guard: `normalize_degree` and `house_spans` were **moved to `utilities/core`**, where the calculation code reaches them without depending on the drawing package — which is why three modules had rewritten them on their own. And the fold hidden in the predicate is gone: there is a single numbering.

---

## Houses that run backwards

Above 66.75° **six systems** (Campanus, horizon, Sunshine, Regiomontanus, Polich/Page, APC) return the cusps in **descending** order; the horizon does so at the equator. Counted over 32,844 charts: 23 systems × 357 half-degree latitudes × 4 times of day.

Read forward, a 6° house measured 354. The twelve invisible wedges of the hit test were painted as near-complete rings, stacked, each with `pointer-events: all` — **a click anywhere answered with the last one drawn**. And the house numbers were centred on the opposite side of the wheel.

Two systems (Polich/Page and Sunshine/alt) do worse: the cusps **cross**, the houses really overlap and no direction makes them tessellate. There each wedge keeps at least one degree, moving only its own end.

Proof end to end, not just the suite: point a longitude at the ring and see which house answers. On Campanus at 70°N a86 answered correctly **1 time in 720**; on the horizon and on APC **0 in 720**. Now ~99%, and every miss is within 0.99° of a cusp.

---

## The composite chart

6.3% of composites at ordinary latitudes produced twelve arcs summing to **1080°** instead of 360 — not a house division.

It is not an unknown case: Solar Fire describes it with the same words and repairs it by bringing the offending cusps to the **long midpoint**; Kepler and Sirius call it "flipping the houses by 180 degrees"; Townley prescribes it in his book. Implemented as they do: one angle is kept at the near midpoint and everything else follows. `house_anchor` chooses which — `auto` (default, like Solar Fire), `ascendant` or `midheaven` (Kepler's two methods).

Five later rounds found that the repair, alone, is not enough:

- **A repair that does not tessellate is not a repair.** Consistency was checked by asking whether the angle–cusp identities survived, never whether the ring was a ring: 9,773 frames out of 148,005 declared a frame with a ring that is not a house division. Now it places, checks, and where the check fails the answer is the midpoints.
- **A cusp opposite another in both parents did not stay opposite.** Two longitudes half a circle apart are the same two longitudes in any order, so an average of the set — which *must* be symmetric for the composite of A and B to be that of B and A — gives both the same answer. No tie-break can repair it: it is derived, as the four angles already were.
- **The angle opens the house of its own cusp, and says so.** On a ring with two cusps on the same longitude the reader answers with the first one met, and a Midheaven that *is* the tenth cusp ended up filed in the fourth.
- **Three arcseconds are not half a turn.** The resultant of a vector average vanishes as the pair approaches half a circle, so the error grows as one over the cosine of half the separation: a 1e-9 tolerance read 8 nanodegrees as a broken identity and turned the whole ring by 180°.

Measured over 258,201 composites — 23 systems, latitudes up to 89°, three anchors, with predicates written independently of the code under test: **angles with a value off their own cusp 2,820 → 0; filed in the wrong house 3,786 → 0.**

Against a86, on real charts and not on this branch: rings that are not house divisions **9.9% → 2.2%** at ordinary latitudes and **19.3% → 8.9%** beyond the polar circle; Midheavens filed outside the tenth, **637 → 0** above the polar circle.

---

## What was NOT fixed here, and is written in the notes

An ordinary **natal** chart can still file an angle in the wrong house: under Sunshine at 74.25°N the cusps from the eighth to the twelfth sit on the same longitude and the Midheaven comes out in the eighth. 68 charts out of 8,280 measured, all under that system. **a86 does the same**, so it is not a regression — but it is the same defect one level down, and it must be fixed where the cusps are computed, not in every factory.

---

## The baselines

They are **included** and regenerating them is idempotent. Worth knowing separately: the suite passed anyway without regenerating them, because `compare_svg_lines` uses `rel_tol=0.5` — on a 300px coordinate it tolerates ±150px — and compares only the numbers, not the text. It let months-old baselines through. Not touched here, but it is the next place to look.

---

## What changes for library consumers

Four things, in order of reach. All in the release notes, with the measurements.

1. **Midpoint composites read differently where a86 was wrong.** An a86 chart reloaded on a87 can read a planet in the opposite house; the two are told apart because a87 records `house_anchor`.
2. **Kerykeion no longer installs a handler on the root logger** — a86 did so as a side effect of a module-level `logging.warning`.
3. **The wheel-only modern chart paints its own background** like the other three (it was the only one not to).
4. **An inverted time range raises** instead of returning a sample outside the requested window.

No other breakage: zero names, signatures, fields or `Literal` members removed or reordered; an a86 payload validates on a87 and vice versa. Over 247 subjects and 448 charts — every house system, ten latitudes, four perspectives, five sidereal modes, four centuries — **no ordinary chart changes by a single number**, and the `kr:` vocabulary is identical across 192 SVGs.

---

## Downstream

- **API v6**: only the pin bump to `6.0.0a87` is needed.
- **Studio**: no code change — it reads `data-kr-node="HouseSector"` and computes no geometry, so with the right wedges its hit test becomes correct on its own. Its 122 SVG baselines must be regenerated in the same train.
- **`6.0.0a87` must be published**: `a86` is already on PyPI and this tree is no longer a86.

---

## What I ask of the review

The branch went through twenty rounds of adversarial review, the last two **with no behavioural defects** over 17,946 randomly generated composites. What the last rounds kept finding were guards that decide without a test behind them, and wrong numbers in comments. The hottest areas, in order:

1. `kerykeion/composite_subject/factory.py` — deeply rewritten, it is where almost all the late defects came out.
2. `kerykeion/charts/utils.py` and `draw_modern.py` — the drawing of the house wedges and their separation.
3. `kerykeion/utilities/core.py` — the functions moved here and the `% 360` trap.

@coderabbitai full review

## Summary by CodeRabbit

* **New Features**
  * Composite charts support configurable house anchoring with `auto`, Ascendant, or Midheaven options.
  * Topocentric ephemeris calculations support observer altitude.
  * Reports and context output include house-frame, fallback, and accessibility details.
  * Added localized astronomical perspective labels and chart accessibility summaries.

* **Bug Fixes**
  * Improved house-sector direction, clickable wedges, numbering, Gauquelin rings, and composite divisions.
  * Corrected degree normalization, boundary formatting, fixed-star naming, and warning behavior.
  * Invalid inverted ephemeris ranges now raise a clear error.

* **Changed**
  * Updated to version 6.0.0a87 and removed the documented “light” theme.


g-battaglia:
@coderabbitai full review

Please review the entire diff of this branch against `alpha/v6`, not just the latest commits — this branch went through twenty review→fix rounds and several defects were introduced by the fixes for earlier ones, so the final state is what matters rather than any single step.

Where to look hardest, in order:

1. `kerykeion/composite_subject/factory.py` — rewritten in depth. `composite_frame` places the twelve cusps on a frame hung from one angle, checks the result is actually a division of the circle into twelve houses, snaps a cusp back opposite its partner where both parents had them opposite, and falls back to plain midpoints where no frame spans the two charts. Please scrutinise the interaction between those four mechanisms, and the direction of the `(source, derived)` pairs.
2. `kerykeion/charts/utils.py` and `kerykeion/charts/draw_modern.py` — house-sector wedges when the cusp ring descends or crosses itself, and the SVG sweep/large-arc pairing that goes with it.
3. `kerykeion/utilities/core.py` — `normalize_degree`, `house_spans` and `get_planet_house` moved here so the calculation code can reach them without importing the drawing package.

Two things worth knowing about the tests: every fix in this branch was mutation-checked (revert the fix, its test must fail alone), and the SVG baseline comparator `compare_svg_lines` uses `rel_tol=0.5` and compares only numbers — so a green suite is weaker evidence than it looks for anything SVG-shaped.

g-battaglia:
@coderabbitai Thanks — eight of your ten findings were real and are fixed in `9e930b55`, each one reproduced by running before being touched and each fix mutation-checked (revert it, its test fails alone).

Fixed: the `house_anchor` value-semantics trio (`__copy__` silently reverted a Midheaven-anchored factory to `auto`; `__eq__`/`__hash__` collapsed two factories that produce different charts); the star name reaching `fixstar_ut` unstripped; the catalog warning repeating on "no conjunction found" rather than "resolved nothing"; the two `zip()` calls in `house_spans` now `strict=True`; the extended baseline script now reading `HOUSE_SYSTEM_NAMES` from the shared matrix in both loops; the release notes describing the surface as additive rather than unchanged; and the composite repair rate unified at one pair in sixteen.

**Two I am declining, with the measurement rather than an opinion:**

**`format_absolute_degrees` (serializer.py:162).** 30.001 degrees IS zero degrees of Taurus, and `"30.00"` is that same sign — not the next one. Rounding *down* cannot leave a sign, because the floor belongs to it; only rounding up can, which is what the guard exists for. Swept over 20,144 values — every sign, both boundaries, plus 20,000 random longitudes — **zero** format outside their own sign's `[floor, floor+30)`.

Your finding did surface something real, though: nothing was testing that function at all. `grep format_absolute_degrees tests/` returned nothing. The sweep is now a test, along with the 149.99687 → `"149.99"` case the docstring cites.

**`translation_strings.py:949`.** The pack is `"CN"`, which is Simplified Chinese, not Traditional — and the same block already uses 个 (Traditional 個) and 运 (Traditional 運). `质心` is correct as it stands; changing it to `質心` would be the inconsistency.

One note on the finding you filed against `composite_subject/factory.py:631`: Codex found the same three methods independently, which is a good sign for both of you.

g-battaglia:
@coderabbitai full review

Ten commits have landed since your last pass (`868a250b` → `7c5c9620`), all from nine rounds of adversarial review by a second reviewer. Please review the **whole diff against `alpha/v6` in its final state** rather than incrementally — several of those commits fix defects that earlier commits on this same branch introduced, so any single step is misleading.

Where the changes concentrate, in order:

1. `kerykeion/composite_subject/factory.py`. The frame logic was rebuilt again. Its current shape: (a) a frame exists only if **both** parents are house divisions running the **same way** (`_rings_admit_a_common_frame`); (b) a cusp the parents put exactly opposite another is snapped back onto it; (c) the ring is placed on the frame, then checked to be a house division, falling back to plain midpoints when it is not; (d) the ring is reconciled with the angle frame by the **arc** each angle stands at from its own cusp — zero being only the special case — with an exact identity outranking an arc, and the anchor breaking the tie when neither is exact. Please scrutinise the interaction of those four, and anything that could make `composite(A,B)` differ from `composite(B,A)`.

2. Derived points. The composite now materialises the four angles regardless of `active_points`, and derives every opposite in `OPPOSITE_PAIRS` from its primary at +180 — closing the pair **from either end**, because a symmetric mean cannot distinguish an antipodal pair from its own reverse.

3. `kerykeion/utilities/core.py`: `get_planet_house` now reads containment from `house_spans` where the twelve are a division, and its exact-on-cusp rule takes the **nearest** cusp rather than the first within tolerance.

4. `kerykeion/charts/utils.py` and `draw_modern.py`: overlapping house wedges are painted widest-first so the narrowest is topmost, matching the reader; ordering uses **exact** spans while overlap detection uses the **drawn** (quantised) ones; and the modern cusp-label crowding policy measures separations direction-independently.

Two things that cost a round each, so you do not repeat them:

- The package in `.venv` is an **editable install pointing at the working tree**, and it wins over `PYTHONPATH` and the cwd. Verify `kerykeion.__file__` before trusting any comparison against another revision.
- Do **not** measure with predicates imported from the code under test. One "zero violations" result in this branch was worthless because the checking script shared the code's own wrong convention.

And as before: `compare_svg_lines` in `tests/core/test_chart_drawer.py` uses `rel_tol=0.5` and returns silently on any non-numeric difference, so a green suite is weak evidence for anything SVG-shaped.

Every fix on this branch is mutation-checked: revert it and its test must fail alone. If you find something, an exact reproducing input matters more than a diagnosis — I verify each finding by running it before touching anything, and two of your ten last time turned out not to reproduce.

g-battaglia:
@coderabbitai full review

g-battaglia:
Two traps that cost the other reviewer a round each, if you run anything:

1. The package in `.venv` is an editable install pointing at the working tree, and it wins over `PYTHONPATH` and the cwd. Check `kerykeion.__file__` before trusting a comparison against another revision.
2. Do not measure with predicates imported from the code under test — one "zero violations" result on this branch was worthless because the script shared the code's own wrong convention.

Also: `compare_svg_lines` in `tests/core/test_chart_drawer.py` uses `rel_tol=0.5` and returns silently on any non-numeric difference, so a green suite is weak evidence for anything SVG-shaped.

g-battaglia:
Four of the five applied in `9d3d7ab1`, each mutation-checked.

The generator one was the real find. Twelve handlers printed `ERROR ...` and carried on, `main()` printed TOTAL CHARTS GENERATED and exited 0 — so a baseline that could not be drawn stayed exactly as it was while the run reported success, and the comparison test then read the stale file. This branch has already been bitten by that shape twice. It exits non-zero now, and it found something on the first run: on an installation whose ephemeris covers only 1550 onwards, the ancient temporal baselines (Constantinople 400AD, Charlemagne 800AD) cannot be drawn at all, and every regeneration of them has been reporting success without touching the files.

Also applied: `zip(..., strict=True)` in `spreading.py`; `__all__` sorted (51 names before and after, none lost); and `house_frame` documented in the skill reference, along with what the anchor is and is not.

**Declining the `East_Point` one**, with the measurement rather than an opinion: there is no populated `east_point` to omit. `East_Point` is not a member of the `AstrologicalPoint` literal — `active_points=["East_Point"]` raises `KerykeionException` — and `AstrologicalSubjectModel` has no `east_point` field. The name survives only as an SVG glyph symbol and inside a filter set. Adding it to `AXES_AND_NODES_IN_READING_ORDER` would advertise a point no chart carries.

PR #255, opened 2026-08-25
fix: the two open points of #253 — the angle that opened a house it was not filed in, and the golden comparison that could not fail

g-battaglia:
Closes the two points left open by #253, then five rounds of code review with fixes (details in the comments).

## 1. The angle opens the house of the cusp it is

At extreme latitudes some systems put several cusps on the same longitude (Sunshine at 74.25 N: 2nd–6th on `316.971024`, bit for bit). `get_planet_house` answered with the lowest one, and an Imum Coeli that **is** the fourth cusp ended up in the third — in the model, in the report, in the context.

**pyswisseph returns the same ring**, so the cusps are not touched: diverging from the reference the library is validated against would cost more than it solves. What changes is *who answers for the angles*: the angle↔cusp identity is recorded where cusps and angles come out of the same call — `houses_ring_with_polar_fallback` for the natal, `angle_house_identities` on the already ayanamsa-shifted ring for the relocated chart, the composite on its own ring — and the consumers that project a point into its **own** ring (house comparison, solar arc at zero arc) read the house from the model. The new `coincident_house_cusps` field declares the zero-width houses (validated: groups of ≥2 houses in 1..12, ascending, no duplicates; derived from the cusps for earlier payloads); report and context show it.

Measured: 424,764 charts per backend + 183,120 per minute, base vs branch — cusps and positions identical, **a single different house**, the angle of the corrected case; zero invariant violations.

## 2. The golden comparison that could not fail

The problem was not `rel_tol=0.5` but three silent `return`s and an escape hatch at ±5% of the line count. A single comparator (`tests/data/compare_svg_lines.py`): structure fatal on both backends; numbers at `rel_tol=0, abs_tol=1e-4` only on libephemeris, with a declared skip elsewhere; missing baseline = failure; regeneration refused on a foreign backend or checkout.

As soon as it could fail it found: golden charts asking GeoNames where they were born (2/6/63/69 failures on an unchanged tree), twenty baselines read by nobody, eleven that no script produced, two writing the same file, nine "unregenerable" ones that all regenerate. Coordinates frozen in `tests/data/golden_places.py`; `tests/data/golden_drive.py` drives every golden test for two gates — hermeticity (network refused, and every reachable baseline must be handed to the comparison) and coverage (every baseline has a reader, with the driver as the only witness in the driven modules). `poe check` runs the gates on the extended kernel too and refuses a kernel narrower than the one requested; `poe regenerate:svg` ends with the eleven test-owned baselines.

## Gates
[...]

PR #256, opened 2026-08-25
fix(timing): the last day of the calendar exits cleanly, and the relocated sidereal Sunshine is the natal chart

g-battaglia:
Closes the two pre-existing issues reported by the #255 review.

## 1. The day after 9999-12-31 does not exist
On the extended kernel (DE441, up to 17191) 9999-12-31 is computable. Two factories let a raw error out for the day **after**:
- `PlanetaryHoursFactory` — `date(9999,12,31) + 1 day` → `OverflowError` (and symmetrically towards 0001-01-01 west of UTC) → now a `KerykeionException` naming the calendar (`_civil_day_beside`);
- `VoidOfCourseMoonFactory` — the Moon's walk past the end of the year → `ValueError: year 10000` from `julian_day_to_utc` → now `KerykeionException` (the single-moment path deliberately excludes `ValueError` from the normalised exceptions, so the wrapping lives in the helper).

The test `test_civil_range_edge_days_raise_clean_exception` demanded that the 31st itself fail: true on the base/medium kernels (coverage), false on the extended one — and there it never reached the two lines that would have caught the defect. It now **probes** the ephemeris for 9999 (a swisseph install passes the tier probe at year 1000 but has no `sepl_96.se1`) and an `@extended` test pins the asymmetry: at 03:00 on 9999-12-31 the planetary hours work (yesterday is needed), at 23:30 they do not; the VoC on 12-01 works, on the 31st at 23:30 it does not.

## 2. Sidereal Sunshine `i` in the relocated chart
Verified on **both** backends: tropical `i` inside the polar circle raises; sidereal `i` returns the ring of `I`. It is the Swiss Ephemeris rule (Makransky → Treindl under the sidereal flag, except in the fixed-epoch modes), and libephemeris mirrors it in `houses_ex`: **nothing to push upstream**. Kerykeion's relocated chart calls the houses in tropical + ayanamsa shift, so it asked for `i`: at 60 N the two constructions are tens of degrees apart, at the pole the refusal was replaced by Porphyry where the natal has the crowded ring. `_house_system_the_reference_casts` asks for `I` exactly in the cases where the reference would.

Tests: `TestSiderealSunshineFollowsTheReference` (60 N, pole, J2000 as a guard against a wrong extension of the rule, polar tropical `i` still replaced), on libephemeris and swisseph. Mutation: without the rule exactly the two LAHIRI tests fail.

## Gates
- libephemeris medium `test:all`: **12,288 passed, 0 failed**.
- libephemeris extended `test:all`: **13,591 passed**; the pre-existing `sun_times` failure is gone. 5 BCE failures remain (`Ancient Greece 500BC …`) that are on `origin/alpha/v6` regardless of this PR: the panel line changed with #254 and the baselines are updated in the not-yet-pushed local commit `dfebf107` — they disappear with its push.
- `test:swe`: the usual 3 pre-existing ones.
- `ruff`, `mypy`, `docs:check` clean.

## Left open, declared
Relocations in the fixed-epoch modes (J2000/J1900/B1950/Mardyks) drift for **every** house system — thousandths of a degree with J2000, up to five hundredths with J1900, the same on both backends: they shift by a flat ayanamsa where the reference applies the epoch's frame. Pre-existing; the J2000 test uses a 0.1° tolerance (Makransky and Treindl are tens of degrees apart) and says so.

https://claude.ai/code/session_01QHgj2emQBET9RL2XayUxf7

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling for dates outside the supported civil range, including planetary hours, lunar ingress, and solar-time calculations.
  - Relocated sidereal Sunshine charts now use reference-compatible house-system calculations.
  - Preserved existing behavior for fixed-epoch modes and non-sidereal charts.

- **Tests**
  - Added coverage for year-9999 boundaries, extended ephemeris ranges, polar relocation scenarios, and sidereal house-system behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,33 @@
 
 ### Fixed
 
+- **Two factories let a raw error out at the end of the civil range.** On the
+  full-range kernel (DE441, to the year 17191) 9999-12-31 is computable, and it is
+  the day AFTER it that does not exist: the planetary day beginning at that
+  evening's sunrise needs a sunrise on 10000-01-01 and overflowed `date`
+  arithmetic with a bare `OverflowError`; the void-of-course Moon walking to its
+  next ingress past the year's end surfaced a bare `ValueError` from the Julian
+  Day conversion. Both refuse with `KerykeionException` now, naming the civil
+  range rather than the ephemeris. The test that guards this had asserted that
+  9999-12-31 itself must fail — true on the base and medium kernels, where the
+  coverage check refuses first, and false on the extended one, where it never
+  reached the two lines that would have caught the real defect. It says the
+  truth at every tier now, and an extended-only test pins the asymmetry: before
+  that day's sunrise the planetary hours work (yesterday exists), after it they
+  cannot.
+
+- **A sidereal Sunshine `i` relocated onto its own birthplace was a different
+  chart.** The reference implementation casts a sidereal `i` (Makransky) as `I`
+  (Treindl) — libephemeris matches it — except in the fixed-epoch modes; the
+  natal chart got that ring by asking with the sidereal flag, while the relocated
+  chart asks tropically and applies the subject's ayanamsa afterwards, so it asked
+  for `i` and got Makransky: at sixty degrees north the two constructions are
+  tens of degrees apart, and inside the polar circle the tropical `i` is refused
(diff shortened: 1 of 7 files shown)


g-battaglia:
@coderabbitai review

PR #249, opened 2026-08-12
feat(cli): optional kerykeion[cli] — the whole library from the terminal (6.0.0a85)

g-battaglia:
## What

Adds an **optional** command-line interface, `kerykeion[cli]`, that exposes the whole library from the terminal — every chart type, analytical technique, sky event, factory and time series — without writing any Python.

**No calculation changed, and no public name moved.** `kerykeion.__all__` is exactly what `a84` exported, every Pydantic schema is byte-identical, and `import kerykeion` is unchanged. The CLI is an `extra`: `pip install kerykeion` without it behaves exactly as before.

Release: **`6.0.0a85`**.

## What you can do from the terminal

```
pip3 install "kerykeion[cli]"
```

- **Charts:** `natal`, `synastry`, `transit`, `composite`, `return`, `progression`, `now`
- **Techniques:** `technique` (profections, firdaria, zodiacal releasing, receptions, horary, midpoints, directions, acg, stars, relocate)
- **Sky events:** `sky` (eclipses, lunations, ingresses, stations, mundane, phenomena, nodes, sun-times, hours, voc)
- **Time series:** `ephemeris`, `transits` (with a pre-flight sampling guard → exit `8`)
- **Subject profiles:** `subject save|show|list|path|verify` (perms `0600` — birth data is PII), reused via `-s <name>`
- **`call`:** a guarded dispatcher reaching any public `Factory.method` in `kerykeion.__all__`

```bash
kerykeion subject save ada --name "Ada Lovelace" --date 1815-12-10 --time 18:00 \
      --lat 51.5074 --lng -0.1278 --tz Europe/London --offline
kerykeion natal -s ada                        # TTY  → ASCII report
kerykeion natal -s ada | jq -r .sun.sign      # pipe → JSON, no ANSI
kerykeion natal -s ada -f svg -o /tmp/ada.svg
kerykeion call DominantsFactory.from_subject -s ada
kerykeion call os.system --param cmd=ls       # refused — os ∉ __all__
```

## Design invariants

- **`import kerykeion` stays free of `typer`.** The console script is static metadata, so the command is installed even without the extra: it then prints an install hint and exits `3` — never a traceback. Enforced in `cli/__init__.py` (lazy import), `kerykeion/__init__.py` (untouched — never imports `kerykeion.cli`), the `walk_packages` exclusion in `test_public_api_surface.py`, and a dedicated subprocess test.
[...]

PR #263, opened 2026-09-02
feat(cli): kerykeion-cli — the whole library from the terminal, as its own distribution behind [cli]

g-battaglia:
## What

A command-line interface for kerykeion, shipped as **its own distribution** — `kerykeion-cli`, built from `cli/` in this repository — which the `kerykeion[cli]` extra installs. It exposes the whole library from the terminal: every chart type, analytical technique, sky event, factory and time series, without writing any Python.

**No calculation changed and no public name moved.** `kerykeion.__all__` is what `a92` exports, every Pydantic schema is byte-identical, and `import kerykeion` never imports the CLI.

> Replaces #249, whose description documented the original a85 design (typer + rich, the CLI inside the library package, an install hint and exit `3`). None of that survived; the branch is the same, the description is now accurate.

## Two distributions, one version

`[project.scripts]` is static metadata, so while the CLI lived inside the library **every** `pip install kerykeion` installed a `kerykeion` command and the `[cli]` extra was a label with nothing behind it. Now:

| | ships | command |
|---|---|---|
| `pip install kerykeion` | the library | none |
| `pip install "kerykeion[cli]"` | library + `kerykeion-cli` | `kerykeion` |
| `uv tool install kerykeion-cli` | both, isolated | `kerykeion` |

`uv` treats the two as a workspace: one lockfile, one `.venv`, one version. The CLI pins the library exactly (`kerykeion==6.0.0a92`) and both are released together. Nothing in the resolver enforces that, so `tests/core/test_cli.py::TestEntryPoint::test_one_version_two_distributions` does, naming the number that drifted. `uv sync` does not install extras, so the CLI is in the `dev` group as an editable member — that is what puts the command on `.venv/bin` for `skill:cli:smoke`.

`poe build:smoke` proves the split from both sides: the library wheel alone renders a chart and declares no console script; the two wheels together serve the command end to end.

## No third-party dependency

typer and rich are gone. The interface is `argparse` over the library, so **kerykeion is the whole dependency of `kerykeion-cli`** and the wheel adds nothing to a user's graph. A command is a plain function whose parameters are `Annotated[T, Opt(...)]` aliases from `options.py`; `parser.py` reads the signature and declares the arguments, so a flag is spelled once. With no dependency to miss there is no install hint and no exit `3`: wherever the command exists, every command works.

`NOTICE` lost the typer/rich block, which stopped being true when the CLI went stdlib-only.

## What you can do from the terminal

```bash
pip3 install "kerykeion[cli]"          # or: uv tool install kerykeion-cli
```

- **Charts:** `natal`, `now`, `synastry`, `transit`, `composite`, `return`, `progression`
- **Analyses:** `aspects`, `dominants`, `moon`, `relationship-score`
- **Techniques:** `technique` (profections, firdaria, zr, receptions, horary, midpoints, directions, acg, heliacal, relocate, nodes, house-comparison, solar-arc, fixed-stars)
- **Sky events:** `sky` (sun-times, hours, phenomena, occultations, voc, eclipses, lunations, ingresses, stations, mundane)
- **Time series:** `ephemeris`, `transits`
- **Subjects:** `subject save|show|list|path|verify` (`0600` in a `0700` store — birth data is PII), reused with `-s <name>`
- **`call`:** a guarded dispatcher reaching any public `Factory.method` in `kerykeion.__all__`
- **`info`, `status [--check]`:** the CLI describes itself and judges its own install

```bash
kerykeion subject save ada --name "Ada Lovelace" --date 1815-12-10 --time 18:00 \
      --lat 51.5074 --lng -0.1278 --tz Europe/London --offline
kerykeion natal -s ada                        # TTY  → ASCII report
kerykeion natal -s ada | jq -r .sun.sign      # pipe → JSON, no ANSI
kerykeion natal -s ada -f svg -o /tmp/ada.svg --theme dark
kerykeion sky ingresses --from 2026-01-01 --to 2026-12-31 --periods
kerykeion call DominantsFactory.from_subject -s ada
kerykeion call os.system --param cmd=ls       # refused — os ∉ __all__
```

## Design invariants

- **`import kerykeion` never imports the CLI.** Enforced by a subprocess test, by `scripts/cli_smoke_check.py`, and now structurally: the CLI is not in the library package at all.
- **Output discipline.** Payloads go to stdout through `sys.stdout` / `Path.write_text`, verbatim: no wrapping, no colour. Format is chosen for you — **text on a TTY, JSON in a pipe** (`-f text|json|xml|svg` forces, `-o file` infers from the suffix). Warnings always go to **stderr**, even under `-f json`, so a piped payload stays clean; `--envelope` folds them in-band for a consumer that only reads stdout, on every command that produces a payload. `-f` is the one format flag: `--json` is gone from `status` and `call`, and `status` follows the same TTY/pipe rule as everything else.
- **Clean errors, never a traceback.** One line on stderr with a classified exit: `0` ok · `1` unexpected · `2` usage · `4` invalid input · `5` KerykeionException · `6` ephemeris · `7` network · `8` sampling ceiling · `9` warnings-as-errors · `130` interrupted.
- **Security is an allowlist.** `call` resolves only names in `kerykeion.__all__`, split on a single `.`, no private members, `getattr_static` so no descriptor runs. `kerykeion call os.system` is refused by an explicit test.
- **Backend-agnostic.** Same in-process backend as the library (libephemeris by default, Swiss Ephemeris optional).

## Also in this branch

- **Simplifications.** `subject save --snapshot` (a cache of the computed subject) is gone: a subject computes in ~40 ms against a second of library start-up, so it saved nothing a user notices while giving a profile a way to go stale — and profiles written by earlier versions still load. The pre-flight sample count is gone too: the factory already counts samples in its constructor, so the CLI only rewords that refusal as exit `8` with the flags to change. `doctor` is now `status --check`.
- **a92 follow-up.** `sky ingresses --periods` and `sky stations --periods` reach the new span queries (sign stays, retrograde spans), with identity tests against the factories.
- **Help reads as a menu**, `-h` everywhere, one line per command. `technique stars` → `technique heliacal` (next to `fixed-stars` the old name said the opposite of what it did). `--major-aspects-only` (the default) dropped; `--all-aspects` stays.

## Tests

`tests/core/test_cli.py` (196 cases, `cli` marker) asserts **identity, not snapshots**: a command's payload `== json.loads(model.model_dump_json())` built from the same inputs, which is immune to ephemeris drift — if the Sun moves, both sides move together. There is deliberately no `regenerate:cli` task. The suite runs offline on base-tier dates, drives `main(argv)` in-process through a small runner that captures the streams, and uses a subprocess only where that cannot prove the point (the real entry point, import isolation).

## Gates

All local, as the project requires — **no CI was added**, `.github/` holds only `FUNDING.yml`.

- `poe check` — green: `ruff`, `mypy` (157 files), `pyright` (0 errors), import graph + fresh-imports, **7578** core tests, **1155** baseline-gate tests, `cli:smoke`, `skill:cli:smoke` (39/39 shell blocks in the agent skill)
- `poe test:cli` — 187 passed
- `poe build:smoke` — both distributions built; library wheel alone renders a chart and carries no command; both wheels serve `--version`, `--help`, a clean exit `4`, `info` and `status --check`

## Review

@coderabbitai full review — please pay particular attention to:

1. **The packaging split**: `cli/pyproject.toml` against the root `pyproject.toml` (workspace members, `[tool.uv.sources]`, the exact pin, the `dev`-group editable, sdist `include` completeness since uv builds the wheel from the sdist).
2. **The argparse layer** `cli/kerykeion_cli/parser.py`: building parsers from function signatures, `Annotated` metadata, `--x/--no-x` pairs, repeatable flags.
3. **The error boundary and exit-code classification** in `cli/kerykeion_cli/errors.py`, and the warnings funnel in `warnings.py` (payload before warnings, `--warnings-as-errors` still winning after a render failure).
4. **`call`'s allowlist** (`registry.py`, `introspect.py`): anything reachable that should not be.
5. Any place the docs, the skill or `NOTICE` still describe the old design.

## Summary by CodeRabbit

* **New Features**
  * Added an optional command-line interface for charts, analyses, techniques, sky events, time series, subject profiles, diagnostics, and public API calls.
  * Added text, JSON, XML, and SVG output options, including file output and pipeline-friendly defaults.
  * Added subject profile management and installation status checks.
  * Added span queries for sky ingresses and stations with `--periods`.
* **Documentation**
  * Added CLI installation guidance, command references, usage recipes, and an AI-agent CLI skill.
* **Changed**
  * The CLI is distributed separately as `kerykeion-cli`.
  * Replaced `doctor` with `status --check`; removed `subject save --snapshot`.


g-battaglia:
@coderabbitai full review

Please review the whole branch, not just the last commit: it replaces #249 and the design changed underneath it. The highest-value areas:

1. **The packaging split** — `cli/pyproject.toml` against the root `pyproject.toml`: workspace members, `[tool.uv.sources]`, the exact `kerykeion==` pin, the editable entry in the `dev` group, and whether the sdist `include` list is complete (uv builds the wheel from the sdist, so an omission ships a broken package).
2. **`cli/kerykeion_cli/parser.py`** — parsers built from function signatures: `Annotated` metadata, `--x/--no-x` pairs, repeatable options, type coercion.
3. **Exit-code classification** (`errors.py`) and the warnings funnel (`warnings.py`): the payload must be written before the warnings, and `--warnings-as-errors` must still win when the renderer itself fails.
4. **`call`'s allowlist** (`registry.py`, `introspect.py`): anything reachable that should not be.
5. Anywhere the docs, the agent skill or `NOTICE` still describe the old typer/rich design or the old module path.

PR #266, opened 2026-09-11
feat(cli): ship a generated kerykeion(1) man page

g-battaglia:
## Why

The CLI had no man page at all: the command tree exists only in argparse, so `man kerykeion` failed on every platform and every install method — `kerykeion -h` was the only terminal documentation.

## What

- **`scripts/generate_cli_manpage.py`** walks the same `build_parser()` behind `--help` and renders deterministic roff into **`cli/man/man1/kerykeion.1`** (committed, generated — never edited by hand).
- The page is *the map*: NAME/SYNOPSIS/DESCRIPTION, the full command tree with per-command synopses and descriptions, global flags, output formats, subject profiles, environment, files, every exit code, examples, and SEE ALSO. Per-command flags are deliberately not duplicated — they stay one `kerykeion <command> --help` away.
- **Packaging**: the `kerykeion-cli` wheel carries the page as `share/man/man1` data (`[tool.hatch.build.targets.wheel.shared-data]`); the sdist includes `/man` so the wheel-from-sdist build keeps it. The library wheel is untouched — it still ships no command and no page.
- **Where it lands**: `pip install` into a system prefix (`/usr/local`, Homebrew) gives `man kerykeion` with no further step. Isolated installs (`uv tool`, a venv) keep the page inside their own prefix; `cli/README.md` and `site/docs/cli.md` document the one-line `man -M` / `MANPATH` addition for those (a wheel cannot safely write outside its prefix).

## Anti-drift (the page is generated, so it must not rot)

- `poe man:generate` regenerates; **`poe man:check`** fails on drift and is now part of `poe check` and `poe quality`.
- `TestManPage` (in `tests/core/test_cli.py`) asserts the committed page matches the parser tree byte-for-byte, renders deterministically, lists every command path (walked from the parser, with a sanity floor), documents every exit code (derived from `ExitCode`), and pins the roff escaping.
- `poe build:smoke` now verifies the page is **installed from the built wheel** and **renders through `man -M`** in the isolated clean room. Locating it goes through the distribution's RECORD: pip materializes data files under the prefix, while uv's cached layered environments keep them in the layer RECORD points into — guessing from `sys.prefix` fails on one of the two.

## Docs

`CHANGELOG.md`, `DEVELOPMENT.md`, `scripts/README.md` (generated-source table), `README.md`, `cli/README.md`, `site/docs/cli.md`.

## Verification

```
poe man:check           OK — 49 commands
poe test:cli            230 passed (6 new)
poe cli:smoke           OK
poe build:smoke         OK — "man page installed and renders"
MANPATH=cli/man man kerykeion   rendered and reviewed by eye
ruff check / format     clean (scripts; tests keep the repo's compact style)
```


g-battaglia:
Addressed both review findings in `068bc1d7`:

- Replaced the 1815 example date with `1900-12-10`, verified end-to-end with `LIBEPHEMERIS_PRECISION=base` (`subject save` and `natal` both exit 0).
- Corrected the generated roff header comments from `.` followed by `"` to proper `.\"` comment lines, and added a regression assertion for the generated prefix.
- Regenerated `cli/man/man1/kerykeion.1`.

Verification: `poe man:check`, all 6 `TestManPage` tests, ruff, manual `man` rendering, and `poe build:smoke` (wheel page installed and renders) all pass.

